PR #6, opened 2022-12-15
Converting Travis CI to Github Actions

j2trumpet:
[Jira ticket](https://jira.cru.org/browse/GT-1881)

The Travis setup was caching npm (which is apparently [automatically included now](https://docs.travis-ci.com/user/caching#npm-cache) anyways), but I've decided to not include that in the GitHub Actions because caching `node_modules` is [not recommended](https://github.com/actions/cache/blob/main/examples.md#node---npm).


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+      # - master
+  pull_request:
+    branches:
+      - jp-gha-conversion
+      - staging
+      # - master
+  workflow_dispatch:
+
+jobs:
+  PreDeployTesting:
```
frett, line 17: I would name this job `build`, if you want to give it a user readable name you can use the `name` property

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+        run: yarn install
+
+      - name: 💅 Prettier check
+        run: yarn prettier:check
+
+      - name: 💨 Lint check
+        run: yarn lint
+
+      - name: 🏗️ Build app
+        run: yarn ng build --prod
```
frett, line 39: I would guess you don't actually need the `ng` in this command, which would follow the same pattern as the other steps

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+
+  Deploy:
+    if: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/staging')}}
```
frett, line 42: I would probably say we want to support doing `workflow_dispatch` for deploys from master & staging as well. That way we can quickly force a redeploy

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+
+      - name: Install AWSCLI
+        run: pip install --user awscli
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@036a4a1ddf2c0e7a782dca6e083c6c53e5d90321
```
frett, line 59: let's use the actual `v1` of this step instead of a specific commit id.

specifying this specific commit was to work around a bug when the github <-> aws integration was first being developed and there were some incompatibilities due to differing release schedules

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Node.js setup
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+
+      - name: 📦 Install dependencies
+        run: yarn install
+
+      - name: 🏗️ Build app
+        run: yarn ng build --prod
```
frett, line 73: again, we probably don't need the `ng` in this step so that it will match the rest of the `yarn` commands

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Copy mobile directory
+        run: cp -r mobile/ dist/mobile/
+
+      - name: Update embed host (staging)
+        if: github.ref == 'refs/heads/staging'
+        run: sed -i "s%$appDomainPlaceholder%$appDomain%g" dist/embed.js
+        env:
+          appDomainPlaceholder: '{appDomain}'
+          appDomain: 'https://stage.knowgod.com'
```
frett, line 89: this should be configured in either the GHA environment or terraform (probably terraform).

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+          aws s3 sync dist/mobile s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache' --content-type 'application/json'
+        env:
+          S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+
+      - name: Deploy prod
+        if: github.ref == 'refs/heads/master'
+        run: echo "Deploy code goes here."
+
+      - name: Deploy summary
+        run: echo "Deployed to the ${{ (github.ref == 'refs/heads/master' && 'prod') || 'stage' }} environment from the ${{ github.ref_name }} branch."
```
frett, line 107: This step is rather pointless, and the message will be buried in the action run log which would pretty quickly show if the deploy succeeded or failed anyways

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+          aws s3 sync dist/knowgod s3://$S3_BUCKET --acl 'public-read' --include '*.html' --cache-control 'no-cache' --delete
+          aws s3 cp dist/embed.js s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache'
+          aws s3 cp dist/example.html s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache'
+          aws s3 sync dist/mobile s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache' --content-type 'application/json'
+        env:
+          S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+
+      - name: Deploy prod
+        if: github.ref == 'refs/heads/master'
+        run: echo "Deploy code goes here."
```
frett, line 104: you shouldn't need separate logic to deploy to production, you should be using the same 2 steps that are used to deploy to stage

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Update embed host (staging)
+        if: github.ref == 'refs/heads/staging'
+        run: sed -i "s%$appDomainPlaceholder%$appDomain%g" dist/embed.js
+        env:
+          appDomainPlaceholder: '{appDomain}'
+          appDomain: 'https://stage.knowgod.com'
+
+      - name: Deploy to s3 (staging)
+        if: github.ref == 'refs/heads/staging'
```
frett, line 92: you shouldn't need a conditional on this step

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+        run: cp embed/embed.js dist/embed.js
+
+      - name: Copy example.html
+        run: cp embed/example.html dist/example.html
+
+      - name: Copy mobile directory
+        run: cp -r mobile/ dist/mobile/
+
+      - name: Update embed host (staging)
+        if: github.ref == 'refs/heads/staging'
```
frett, line 85: you shouldn't need a conditional on this step because it should be used for deploying both to staging and production

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: 💅 Prettier check
+        run: yarn prettier:check
+
+      - name: 💨 Lint check
+        run: yarn lint
+
+      - name: 🏗️ Build app
+        run: yarn ng build --prod
+
+  Deploy:
```
frett, line 41: name this `deploy` and use a `name` property to define the human readable name.

The pattern we've used everywhere else in GHA for job names is lowercase w/ underscores

frett:
you will also want to delete the `.travis.yml` file so that it quits triggering travis-ci

.github/workflows/node.js.yml
j2trumpet: I'll look up how to configure that.

.github/workflows/node.js.yml
j2trumpet: @frett I caught that too and was already planning to use `${{ secrets.AWS_S3_BUCKET }}` but how would I add `https://` to the beginning? Would `https://${{ secrets.AWS_S3_BUCKET }}` work?

.github/workflows/node.js.yml
j2trumpet: This is essentially a `console.log` for testing. It'll be removed before going live.

.github/workflows/node.js.yml
j2trumpet: @frett Once I switch the `Update embed host: env: appDomain` to use the bucket secret, that block can be used for both staging and prod deployment. But the "Deploy to s3" block needs different caching values between the two environments. Do you have a way to dynamically swap those? I agree it could improve the maintainability of the code.

j2trumpet:
> you will also want to delete the `.travis.yml` file so that it quits triggering travis-ci

I figure that will be the last step once the GHA is confirmed working and doing everything Travis has been.

.github/workflows/node.js.yml
frett: don't use `AWS_S3_BUCKET` as the secret, there is no guarantee that the bucket name will match the hostname for the actual site

.github/workflows/node.js.yml
frett: we should probably make the caching values the same, I wonder if that even matters though because our usually S3 website config involves using CloudFront as a caching layer, which would mean S3 cache times are meaningless

frett: yeah, we are using CloudFront, which would be doing our caching for us

frett:
@j2trumpet well the current travis script won't work anymore now because we recreated the `know_god_web` repo which invalidated the encrypted secret in the script.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,109 @@
...
+
+on:
+  push:
+    branches:
+      - jp-gha-conversion
+      - staging
+      # - master
+  pull_request:
+    branches:
+      - jp-gha-conversion
```
frett, line 11: this can be removed already and won't affect any testing you are doing

frett: you are never opening up another pull request to merge code into the `jp-gha-conversion` branch

.github/workflows/node.js.yml
j2trumpet: But isn't that what I'm doing (with the code you gave me) in the "Deploy to s3" step? I double checked the values assigned in Terraform and they're `staging.knowgod.com` and `knowgod.com`.

Side note: I had to add "s3://" to each line of `run` to get that step to work.

.github/workflows/node.js.yml
j2trumpet: So do I need to specify caching values here at all? Should I take out all the `cache-control` options and values?

.github/workflows/node.js.yml
frett: that is for the next step, what this step is doing is updating the `embed.js` to put in the web url you are serving the site from, which is not always going to be the same thing as the s3 bucket name

frett: this step actually has nothing to do with AWS or S3, it just is preparing `embed.js` to be used on the web

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Deploy to s3 (staging)
+        if: github.ref == 'refs/heads/staging'
+        run: |
+          aws s3 sync dist/knowgod s3://$S3_BUCKET --acl 'public-read' --exclude '*.html' --cache-control 'public, max-age=31536000' --delete
+          aws s3 sync dist/knowgod s3://$S3_BUCKET --acl 'public-read' --include '*.html' --cache-control 'no-cache' --delete
+          aws s3 cp dist/embed.js s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache'
+          aws s3 cp dist/example.html s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache'
+          aws s3 sync dist/mobile s3://$S3_BUCKET --acl 'public-read' --cache-control 'no-cache' --content-type 'application/json'
+        env:
+          S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
```
frett, line 102: so a better way to do this instead of having to put `s3://` on every line is to just include the `s3://` when you are defining the `S3_BUCKET` environment variable

.github/workflows/node.js.yml
frett: let's use the cache values from production, I think it's just setting a different cache setting for index.html

frett: I did research and CloudFront by default utilizes the origin (S3 in this case) cache headers to determine how long it caches data.

If we are concerned about delay between merging a commit and seeing the results live on the stage environment we should probably trigger a cache invalidation in CloudFront after we upload everything to S3. But that will be outside the scope of this PR and could be a future optimization

.github/workflows/node.js.yml
j2trumpet: Should I go ahead and update that in Terraform or not worry about it for now?

frett: don't do that in terraform, do that in this workflow.

For reference you can customize the generation of environment variables like this: https://github.com/CruGlobal/android-gto-support/blob/master/.github/workflows/build.yml#L220

.github/workflows/node.js.yml
frett: relevant CloudFront documentation: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html

.github/workflows/node.js.yml
j2trumpet: I get that the steps aren't directly related, but back to my first question, how should I concatenate the `https://` to the front of `${{ secrets.HOST_NAME }}`?

frett: `https://${{ secrets.HOST_NAME }}`

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: 💨 Lint check
+        run: yarn lint
+
+      - name: 🏗️ Build app
+        run: yarn build --prod
+
+  deploy:
+    name: Deploy
+    if: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/jp-gha-conversion')}}
```
frett, line 43: still need to update this conditional to trigger on `workflow_dispatch` for staging and master, and remove your testing branch.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,95 @@
+name: Node.js CI
+
+on:
+  push:
+    branches:
+      - jp-gha-conversion
+      - staging
+      # - master
```
frett, line 8: remove your testing branch and enable `master`

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,95 @@
...
+  pull_request:
+    branches:
+      - staging
+      # - master
```
frett, line 12: you will need to enable master so the build/check step runs on PRs

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/checkout@v3
+
+      - name: Install AWSCLI
+        run: pip install --user awscli
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_ARN }}
```
frett, line 63: Personally I would move this step right before the `Deploy to s3` step, that will keep the scope of what steps actually have access to AWS credentials more limited

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,95 @@
+name: Node.js CI
```
frett, line 1: let's update this name to be something that more relates to what this workflow is doing

potential options:
- Build
- Build & Deploy
- Build knowgod.com
- (something else that makes sense)

frett (approved):

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,94 @@
...
+  push:
+    branches:
+      - staging
+      - master
+  pull_request:
+    branches:
+      - staging
+      - master
+  workflow_dispatch:
+
```
frett, line 13: include this `concurrency` section: https://github.com/CruGlobal/godtools-android/blob/develop/.github/workflows/build.yml#L9-L11

it basically means that only 1 action can run at a time and it will cancel the previous if it hasn't completed yet

PR #32, opened 2023-04-19
Adding on staging workflow

dr-bizz:
Created a GHA workflow that merges PR's and the `main` branch into staging on pushes if the label `On Staging` is added.


dr-bizz:
I didn't realize that we didn't need this GHA as Cru has a GH app which goes this for us
Also, I needed to updated the branch from `main` to `master`

PR #34, opened 2023-04-20
WIP - Gt 1922 implement new shared parser

dr-bizz:
## Description
Removing the old XML Parser and Integrating Cru's new Shard Parser.

## Changes
- Recreated all the components in a new `page` folder called `new-page`.
- The old `page` components have been kept so we can check to ensure no functionality has been missed. You can view the old pages by /old/en/[book-name]/[page-#]
- Created tests for all components & rebuilt pages.
- There are some differences between the Shared Parser and the old XML parser.
    - Resources are loaded in from two places, the manifest and then also the API.
- To understand the type of content, we run the instance of the class through an IF statement.
- Lottie had to be upgraded to run on Angular 13.
- Tips, Spacer, Accoridan, Multiselect & Card will be developed at a later date.

## On Go live
- Open a New PR to remove the button type.


src/app/new-page/component/content-fallback/content-fallback.component.ts
```diff
@@ -39,9 +36,7 @@ export class ContentFallbackNewComponent implements OnChanges {
               changes['item'].currentValue !== changes['item'].previousValue
             ) {
               this.ready = false;
-              this.content = null;
-              // PIZZA
-              // this.fallback = this.item as KgwContentComplexTypeFallback;
+              this.content = this.item;
```
frett, line 39: the fallback component is fully handled within the xml parser itself. So once the new parser is live, this component can be removed

package.json
```diff
@@ -23,6 +22,7 @@
     "@angular/platform-browser": "^13.4.0",
     "@angular/platform-browser-dynamic": "^13.4.0",
     "@angular/router": "^13.4.0",
+    "@cruglobal/godtools-shared": "0.9.0-SNAPSHOT.1267",
```
frett, line 25: you can switch this to `0.9.0` now

src/app/services/xml-parser-service/xmp-parser.service.ts
```diff
@@ -0,0 +1,179 @@
...
+export type Page = org.cru.godtools.shared.tool.parser.model.page.Page;
+export type CallToAction =
+  org.cru.godtools.shared.tool.parser.model.tract.CallToAction;
+export type Header = org.cru.godtools.shared.tool.parser.model.tract.Header;
+export type Hero = org.cru.godtools.shared.tool.parser.model.tract.Hero;
+export type Modal = org.cru.godtools.shared.tool.parser.model.tract.Modal;
+export type TractPage =
+  org.cru.godtools.shared.tool.parser.model.tract.TractPage;
+export type Card =
+  org.cru.godtools.shared.tool.parser.model.tract.TractPage.Card;
```
frett, line 20: so, I would probably export this as `TractPageCard` or put it under the `TractPage` namespace. This `Card` type only shows up as a top level element within a `TractPage`, and shouldn't be confused with the newer `Card` content element that can show up anywhere.

src/app/services/xml-parser-service/xmp-parser.service.ts
```diff
@@ -0,0 +1,179 @@
...
+export type EventId = org.cru.godtools.shared.tool.parser.model.EventId;
+export type Form = org.cru.godtools.shared.tool.parser.model.Form;
+export type Input = org.cru.godtools.shared.tool.parser.model.Input;
+export type Link = org.cru.godtools.shared.tool.parser.model.Link;
+
+export type Spacer = org.cru.godtools.shared.tool.parser.model.Spacer;
+export type Tabs = org.cru.godtools.shared.tool.parser.model.Tabs;
+export type Tab = org.cru.godtools.shared.tool.parser.model.Tabs.Tab;
+export type Accordion = org.cru.godtools.shared.tool.parser.model.Accordion;
+export type ModalCard = org.cru.godtools.shared.tool.parser.model.Card;
```
frett, line 38: connecting to my previous comment, this probably should be what you export as `Card` from here, since it's a standard `Card` that can be used anywhere

src/app/services/xml-parser-service/xmp-parser.service.ts
frett: some of this is just left over artifacts of how the spec has developed, There are 5 models that are only present on `TractPages` and not part of the common content elements. `Header`, `Hero`, `TractPage.Card`, `Modal`, and `CallToAction`.

frett: and then there are a few types are child types for single elements only. `Tabs.Tab`, `Accordion.Section`, `Flow.Item`, `Multiselect.Option`

src/app/new-page/component/training-tip/training-tip.component.html
```diff
@@ -0,0 +1,2 @@
+<ng-container *ngIf="ready && (visible$ | async)"></ng-container>
+<!-- Training tip content goes here -->
```
frett, line 2: as it stands right now, there will never be any training tip rendering happening on knowgod.com. So you can probably get rid of the `<ng-container />` here

yarn.lock
```diff
@@ -1273,6 +1273,14 @@
   resolved "https://registry.yarnpkg.com/@colors/colors/-/colors-1.5.0.tgz#bb504579c1cae923e6576a4f5da43d25f97bdbd9"
   integrity sha512-ooWCrlZP11i8GImSjTHYHLkvFDP48nS4+204nGb1RiX/WXYHmJA2III9/e2DWVabCESdW7hBAEzHRqUn9OUVvQ==
 
+<<<<<<< HEAD
```
frett, line 1276: hmm, you have a merge conflict marker added to the `yarn.lock` file, you might need to regenerate the yarn.lock file to fix it

src/app/new-page/component/content-animation/content-animation.component.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    this.ready = true;
+  }
+
+  private awaitAnimEvent(): void {
+    this.pageService.contentEvent$
+      .pipe(takeUntil(this._unsubscribeAll))
+      .subscribe((event) => {
+        if (this.animation.attributes.playListeners) {
+          const tEvents = this.animation.attributes.playListeners.split(' ');
```
frett, line 100: are you having to convert the array of `EventIds` into a space separate list to just re-split them here?

src/app/services/xml-parser-service/xmp-parser.service.ts
```diff
@@ -0,0 +1,179 @@
...
+  ) {
+    console.log('CONTENT: Link');
+    return 'link';
+  } else if (
+    content instanceof org.cru.godtools.shared.tool.parser.model.Tabs
+  ) {
+    console.log('CONTENT: Tabs');
+    return 'tabs';
+  } else if (
+    content instanceof org.cru.godtools.shared.tool.parser.model.Tabs.Tab
```
frett, line 96: I don't believe this case will ever return true, `Tab` doesn't (shouldn't) extend the `Content` type, which means it should never be an element in a `Content` array.

src/app/new-page/component/card/card.component.html
```diff
@@ -0,0 +1,20 @@
+<div
+  attr.id="card-{{ cardIndex }}"
+  class="cardContent"
+  *ngIf="ready"
+  [hidden]="card.isHidden"
+  [ngClass]="{
+    hasHeading: !(card.listeners.length && card.dismissListeners.length)
```
frett, line 7: hmm, this seems weird that there is a `hasHeading` class that is dependent on if `listeners`/`dismissListeners` are defined.

src/app/new-page/component/card/card.component.html
```diff
@@ -0,0 +1,20 @@
...
+  }"
+>
+  <h4
+    *ngIf="!(card.listeners.length && card.dismissListeners.length)"
```
frett, line 11: hmm, I'm curious why we are hiding the card header if there are `listeners` or `dismissListeners` defined

src/app/new-page/component/card/card.component.ts
```diff
@@ -0,0 +1,76 @@
...
+  Content
+} from 'src/app/services/xml-parser-service/xmp-parser.service';
+@Component({
+  selector: 'app-page-new-card',
+  templateUrl: './card.component.html',
+  styleUrls: ['./card.component.css']
+})
+export class CardNewComponent implements OnChanges {
+  @Input() card: Card;
+  @Input() cardIndex: number;
```
frett, line 23: There is a `card.position` property that would give you the position of this card so you didn't need to inject an index when creating the component.

src/app/new-page/component/content-video/content-video.component.html
```diff
@@ -0,0 +1,7 @@
+<ng-container *ngIf="ready && provider === 'YOUTUBE' && videoId">
```
frett, line 1: So, the `provider` and `videoId` checks aren't necessary with the new parser.

If the video is not a youtube video or is missing a videoId, it will not be included in the content array returned to you.

frett: of course it doesn't hurt anything leaving these checks in place

src/app/services/xml-parser-service/xmp-parser.service.ts
dr-bizz: That's a good point. I've renamed it.

src/app/new-page/component/training-tip/training-tip.component.html
dr-bizz: Gotcha, if that's the case, we can remove the entire component.

yarn.lock
dr-bizz: Added gitIgnore to ignore all yarn files.

src/app/new-page/component/content-animation/content-animation.component.ts
dr-bizz: I reworked this page. Somehow I missed testing/building this component with the new parser. I've rebuilt this entire page to work with the Parser now.

src/app/services/xml-parser-service/xmp-parser.service.ts
dr-bizz: You're right; we don't have a `tab` content type. This must have been left in after testing and building the tab component. I've removed it.

src/app/new-page/component/card/card.component.ts
dr-bizz: I don't use the input`cardIndex` anymore, as I use position. I've removed `cardIndex`.

src/app/new-page/component/content-video/content-video.component.html
dr-bizz: I assume the likelihood of us moving away from Youtube to another video platform is very low. If that's the case, I'll remove the checks.

frett: @dr-bizz if we ever do add a different hosting platform, we would have to deal with parsing it in the parser first, and it would be locked behind a feature flag.

This means you wouldn't get the different video types until you are ready to handle them

src/app/new-page/component/card/card.component.html
frett: @dr-bizz I'm assuming this is the same behavior that the previous renderer was doing. In which case we don't need to address this with the PR, but it's something we should look into after the fact

src/app/new-page/component/content-button/content-button.component.ts
```diff
@@ -0,0 +1,102 @@
...
+      this.url = this.toAbsoluteUrl(this.button.url);
+    }
+    if (isEventType) {
+      this.type = 'event';
+      this.events = this.button.events;
+    }
+    this.ready = true;
+  }
+
+  private toAbsoluteUrl(url: string): string {
```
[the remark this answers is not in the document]
frett: yeah, I just confirmed that when we parse a URL we make it absolute, so you won't need this logic

src/app/new-page/component/content-button/content-button.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    const isEventType = !!this.button.events?.length;
+
+    if (isUrlType) {
+      this.type = 'url';
+      this.url = this.toAbsoluteUrl(this.button.url);
+    }
+    if (isEventType) {
+      this.type = 'event';
+      this.events = this.button.events;
+    }
```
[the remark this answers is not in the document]
frett: Removing the usage of type might make sense to open as a followup jira ticket and not address it in this PR

src/app/new-page/component/content-animation/content-animation.component.ts
```diff
@@ -0,0 +1,133 @@
...
+  onAnimationClick(): void {
+    if (this.animation.events) {
+      let action = '';
+      this.animation.events.forEach((event, idx) => {
+        const value = event?.namespace
+          ? `${event.namespace}:${event.name}`
+          : event.name;
+        action += idx ? ` ${value}` : value;
+      });
+      this.pageService.formAction(action);
```
frett, line 69: we will definitely need to revisit how events are handled for knowgod.com. This will probably make the most sense to do when we are adding support for state events. I'll create a JIRA ticket for this

frett: see: GT-2034

src/app/new-page/page.component.ts
```diff
@@ -0,0 +1,747 @@
...
+      const translationid = item.id;
+      const fileName =
+        APIURL.GET_XML_FILES_FOR_MANIFEST + translationid + '/' + manifestName;
+      this.pullParserFactory.setOrigin(fileName);
+      const config = XmlParser.ParserConfig.createParserConfig()
+        .withLegacyWebImageResources(true)
+        .withSupportedFeatures([
+          XmlParser.ParserConfig.Companion.FEATURE_ANIMATION,
+          XmlParser.ParserConfig.Companion.FEATURE_CONTENT_CARD,
+          XmlParser.ParserConfig.Companion.FEATURE_MULTISELECT
```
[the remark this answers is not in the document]
frett: it is also possible to declare content such as:
```xml
<content:text requires-feature="multiselect">Make a selection below</content:text>
```
which is only returned to you if you declare that you actually support multiselects

src/app/new-page/page.component.ts
```diff
@@ -0,0 +1,747 @@
...
+
+          // Loop through and get all resources.
+          this._pageBookIndex.included.forEach((resource) => {
+            const { attributes, type } = resource;
+            if (type === 'attachment') {
+              this.pageService.addAttachment(
+                attributes['file-file-name'],
+                attributes.file
+              );
+              if (!attributes['is-zipped'] === true) {
```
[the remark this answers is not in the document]
frett: But this logic looping through attachments will probably completely go away when we switch to granular distribution

src/app/new-page/component/tract-page/tract-page.component.html
```diff
@@ -0,0 +1,91 @@
...
+      *ngIf="header && (isModal$ | async) === false"
+      [header]="header"
+    ></app-page-new-header>
+    <app-page-new-hero
+      *ngIf="hero && (isModal$ | async) === false"
+      [hero]="hero"
+    ></app-page-new-hero>
+    <ng-container *ngIf="cards && cards.length && (isModal$ | async) === false">
+      <app-page-new-card
+        *ngFor="let card of cards; index as cardIndex"
```
frett, line 21: I'm guessing you don't need the `index as cardIndex` here

src/app/new-page/component/content-repeater/content-repeater.component.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  private init(): void {
+    if (this.items?.length) {
+      this.items.forEach((content) => {
+        if (content['content']) {
+          const hasInputChildren = content['content'].filter(
+            (c) => ContentParser(c) === 'input'
+          );
+          if (hasInputChildren?.length) {
+            this.content.push({ type: 'form', content: content['content'] });
```
frett, line 58: I'm curious about this, what's the need of pushing a form content element here?

src/app/new-page/component/content-animation/content-animation.component.ts
frett: also see: GT-2035 regarding what clicking different content elements should do

src/app/new-page/component/content-button/content-button.component.ts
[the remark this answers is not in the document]
frett: see: GT-2035

src/app/new-page/component/card/card.component.html
dr-bizz: @frett Yes, it was. I couldn't find a place where the card had Listeners and dismissListeners to see why it wouldn't show the header.
I've removed the functionality as it's been there since the very beginning and seems to have no purpose.

src/app/new-page/component/content-video/content-video.component.html
dr-bizz: Awesome! That's a great to know.

src/app/new-page/component/content-button/content-button.component.ts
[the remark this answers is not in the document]
dr-bizz: I agree. Let's open a New PR to remove the button type.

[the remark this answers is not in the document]
dr-bizz: commented this in description to remember

src/app/new-page/component/content-button/content-button.component.ts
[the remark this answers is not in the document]
dr-bizz: Removed `toAbsoluteUrl`

src/app/new-page/component/content-animation/content-animation.component.ts
dr-bizz: Thanks for making a ticket; we should definitely re-code this once we have state set up.

src/app/new-page/page.component.ts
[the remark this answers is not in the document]
dr-bizz: Thanks, I must have left that in while testing.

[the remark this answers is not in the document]
dr-bizz: This is good to know. Is that documented on the Mobile admin tools so users know they can do that?

src/app/new-page/component/content-animation/content-animation.component.ts
```diff
@@ -0,0 +1,133 @@
...
+        if (!this.anmViewItem) return;
+        const playListeners = this.animation.playListeners.filter(
+          (listener) => listener.name === event
+        ).length;
+        const stopListeners = this.animation.stopListeners.filter(
+          (listener) => listener.name === event
+        ).length;
+        const events = this.animation.events.filter(
+          (e) => e.name === event
+        ).length;
```
frett, line 120: so you shouldn't be using the `events` property here as listeners, the `events` property is just events that are triggered when the animation is clicked on.

src/app/new-page/component/content-animation/content-animation.component.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        if (playListeners || stopListeners) {
+          if (events) {
+            this.anmViewItem.togglePause();
+          } else if (playListeners) {
+            this.anmViewItem.play();
+          } else if (stopListeners) {
+            this.anmViewItem.stop();
```
frett, line 128: I think you would want to `pause()` the animation instead of `.stop()`, if lottie supports that differentiation on the web

frett: here is how I'm doing play/stop listeners on Android: https://github.com/CruGlobal/godtools-android/blob/develop/ui/base-tool/src/main/kotlin/org/cru/godtools/base/tool/ui/controller/AnimationController.kt#L38-L45

src/app/new-page/page.component.ts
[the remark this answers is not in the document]
dr-bizz: Yeah, here is my workaround for not being able to fetch all the images from the manifest.

src/app/new-page/page.component.ts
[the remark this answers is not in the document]
frett: it's not documented in the tools, but Aaron and Jill both know about that functionality and regularly use it. Unfortunately the previous knowgod.com parser didn't know about the `required-features` attribute and how to respect it.

src/app/new-page/component/content-repeater/content-repeater.component.ts
dr-bizz: From memory, it was because `content['content']` would return a type `Content[]` and not `Form`. So to check if it's a form, I check if there is an Input as a child.

frett: hmm, so technically inputs don't need to be direct children of the form in the content model, although I believe currently they are always direct children.

src/app/new-page/component/content-animation/content-animation.component.ts
dr-bizz: Lottie supports pause and stop, so I've changed it to pause. I didn't think there was a difference. However, Lottie Web doesn't support `resume`.

frett: if I recall, pause maintains the current playback position, stop resets it to the beginning

src/app/new-page/component/content-animation/content-animation.component.ts
dr-bizz: Yes, this function is watching for all events, `play` `stop` and clicks on the animation. which is why I look at the `this.animation.events` as well as the `playListeners` and `stopListeners`.

frett: ah, in that case you shouldn't be doing this below:
```typescript
if (events) {
  this.anmViewItem.togglePause()
}
```

just because an animation can trigger events when it's clicked doesn't mean that it should start/stop playback

src/app/new-page/page.component.ts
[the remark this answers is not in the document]
dr-bizz: Gotcha

src/app/new-page/component/tract-page/tract-page.component.ts
```diff
@@ -0,0 +1,285 @@
...
+  private init(): void {
+    this.pageService.setPageOrder(this.order, this.totalPages);
+    this.pageService.modalHidden();
+    this.pageService.formHidden();
+    this.header = this._page.header || null;
+    this.hasPageHeader = !!this._page.header?.title?.text;
+    this.hero = this._page.hero || null;
+    this.cards = this._page.cards || [];
+    this.modal = this._page.modals ? this._page.modals[0] : null;
+    this.callToAction = !!(
```
canac, line 271: Shouldn't need the `!!` here

src/app/new-page/component/tract-page/tract-page.component.ts
```diff
@@ -0,0 +1,285 @@
...
+      }, 0);
+    } else {
+      this.pageService.contentEvent(functionName);
+    }
+  }
+
+  private setHiddenCardToShow(): void {
+    this.cards
+      .filter((card) => this._cardsHiddenOnFormAction.includes(card.position))
+      .map((card) => {
```
canac, line 249: Since you're not returning anything, `.forEach` would be more appropriate

src/app/new-page/component/tract-page/tract-page.component.ts
```diff
@@ -0,0 +1,285 @@
...
+        (card as any).isHidden = false;
+      });
+  }
+
+  private setShownCardToHidden(): void {
+    (
+      this.cards.find(
+        (card) => card.position === this._cardShownOnFormAction
+      ) as any
+    ).isHidden = true;
```
canac, line 259: You might want to handle the case where `.find` doesn't find anything and returns undefined

src/app/new-page/component/tract-page/tract-page.component.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.pageService.setPageOrder(this.order, this.totalPages);
+    this.pageService.modalHidden();
+    this.pageService.formHidden();
+    this.header = this._page.header || null;
+    this.hasPageHeader = !!this._page.header?.title?.text;
+    this.hero = this._page.hero || null;
+    this.cards = this._page.cards || [];
+    this.modal = this._page.modals ? this._page.modals[0] : null;
+    this.callToAction = !!(
+      this.page.callToAction?.label?.text || this.page.callToAction?.tip
```
frett, line 272: you probably don't need the check for a `tip` since we aren't supporting any tip logic on the web

src/app/new-page/component/content-animation/content-animation.component.ts
dr-bizz: Hmm, I guess there isn't a strict naming convention for the events, so I can't program if an event is a play or stop event. I will remove the event functionality

frett (commented):
I'm good with the current state functionally, I'll approve once QA finishes

frett (approved):
This has passed QA, so I'm approving it

PR #43, opened 2023-05-10
Gt 1759 support accordion functionality

dr-bizz:
Added accordion files.
Added the `Section` type in the XML parser.
Integrated the Shared Parser accordion.
Added 4 files.

This is hard to test without https://github.com/CruGlobal/know-god-web/pull/46, which this PR can wait for PR 46 to be merged.


dr-bizz:
@frett This PR is ready for testing, so I've asked for your review, but this is hard to test on the web without hooking the app up to the staging database. So you can wait until https://github.com/CruGlobal/know-god-web/pull/46 is merged into the main before testing.

src/app/services/xml-parser-service/xmp-parser.service.ts
```diff
@@ -102,8 +102,14 @@ export const ContentParser = (content: any): string => {
...
   ) {
-    // console.log('CONTENT: Accordion');
+    console.log('CONTENT: Accordion');
     return 'accordion';
+  } else if (
+    content instanceof
+    org.cru.godtools.shared.tool.parser.model.Accordion.Section
+  ) {
+    console.log('CONTENT: Section');
+    return 'section';
```
frett, line 112: I don't know if you need to include this here, an `Accordion.Section` doesn't extend the `Content` interface, so it should never show up in a `Content` array. It should only be used in the `accordion.sections` array.

dr-bizz: I've removed this type

PR #78, opened 2023-06-14
GT-2021, GT-2022, GT-1429 Support for  Multiselects, Flows and new emoji resource.

dr-bizz:
Added support for Multiselect, Flow, Flow Cards & upgraded emoji survey to use new Multiselect.

## Changes
- Added support for Multiselect, Flow, and Flow Cards (known as ContentCards).
- Added tests for Multiselect & Flow
- Editing styles for hero header and image to fix margins.

### DIFF
--- a/src/app/new-page/component/content-card/content-card.component.html
+++ b/src/app/new-page/component/content-card/content-card.component.html
@@ -1,0 +1,17 @@
+<ng-container *ngIf="ready">
+  <div
+    *ngIf="!card.isClickable"
+    class="card"
+    [ngStyle]="{ 'background-color': background }"
+  >
+    <app-content-new-repeater [items]="contents"></app-content-new-repeater>
+  </div>
+  <a
+    *ngIf="card.isClickable"
+    class="card clickable"
+    (click)="eventClick()"
+    [ngStyle]="{ 'background-color': background }"
+  >
+    <app-content-new-repeater [items]="contents"></app-content-new-repeater>
+  </a>
+</ng-container>
--- a/src/app/new-page/component/content-flow-item/content-flow-item.component.ts
+++ b/src/app/new-page/component/content-flow-item/content-flow-item.component.ts
@@ -1,0 +1,63 @@
+import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import {
+  Content,
+  FlowItem,
+  FlowWatcher
+} from 'src/app/services/xml-parser-service/xmp-parser.service';
+import { PageService } from '../../service/page-service.service';
+
+@Component({
+  selector: 'app-content-new-flow-item',
+  templateUrl: './content-flow-item.component.html',
+  styleUrls: ['./content-flow-item.component.css']
+})
+export class ContentFlowItemNewComponent implements OnChanges {
+  // eslint-disable-next-line @angular-eslint/no-input-rename
+  @Input() item: FlowItem;
+  contents: Content[];
+  ready: boolean;
+  state: any;
+  isHidden: boolean;
+  isHiddenWatcher: FlowWatcher;
+
+  constructor(private pageService: PageService) {
+    this.state = this.pageService.parserState();
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    for (const propName in changes) {
+      if (changes.hasOwnProperty(propName)) {
+        switch (propName) {
+          case 'item': {
+            if (
+              !changes['item'].previousValue ||
+              changes['item'].currentValue !== changes['item'].previousValue
+            ) {
+              this.ready = false;
+              this.init();
+            }
+          }
+        }
+      }
+    }
+  }
+
+  ngOnDestroy(): void {
+    this.isHiddenWatcher.close();
+  }
+
+  private init(): void {
+    if (this.isHiddenWatcher) this.isHiddenWatcher.close();
+
+    this.isHiddenWatcher = this.item.watchIsGone(
+      this.state,
+      (value) => (this.isHidden = value)
+    );
+    const contents: Content[] = [];
+    this.item.content.forEach((content) =>
+      content ? contents.push(content) : null
+    );
+    this.contents = contents;
+    this.ready = true;
+  }
+}
--- a/src/app/new-page/component/content-flow/content-flow.component.ts
+++ b/src/app/new-page/component/content-flow/content-flow.component.ts
@@ -1,0 +1,50 @@
+import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Flow } from 'src/app/services/xml-parser-service/xmp-parser.service';
+import { PageService } from '../../service/page-service.service';
+
+@Component({
+  selector: 'app-content-new-flow',
+  templateUrl: './content-flow.component.html',
+  styleUrls: ['./content-flow.component.css']
+})
+export class ContentFlowNewComponent implements OnChanges {
+  // eslint-disable-next-line @angular-eslint/no-input-rename
+  @Input() item: Flow;
+
+  flow: Flow;
+  items: any[];
+  ready: boolean;
+  state: any;
+
+  constructor(private pageService: PageService) {
+    this.state = this.pageService.parserState();
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    for (const propName in changes) {
+      if (changes.hasOwnProperty(propName)) {
+        switch (propName) {
+          case 'item': {
+            if (
+              !changes['item'].previousValue ||
+              changes['item'].currentValue !== changes['item'].previousValue
+            ) {
+              this.ready = false;
+              this.flow = this.item;
+              this.init();
+            }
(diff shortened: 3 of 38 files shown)


caleballdrin (approved):

src/app/new-page/component/content-multiselect-option/content-multiselect-option.component.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+  }
+
+  onClick() {
+    this.option.toggleSelected(this.state);
+  }
+
+  private init(): void {
+    this.option.watchIsSelected(this.state, (value) => (this.selected = value));
```
frett, line 50: this returns a `FlowWatcher` object that has a `close()` method that should be called when you no longer want notifications about the selected state.

You will probably want to capture this in class property that can be properly disposed of when this component is destroyed

src/app/new-page/component/content-multiselect/content-multiselect.component.css
```diff
@@ -0,0 +1,38 @@
...
+
+}
+
+.multiselect[data-columns="5"] app-content-new-multiselect-option {
+  flex: 0 1 calc(20% - 12px);
+}
+
+.multiselect[data-columns="6"] app-content-new-multiselect-option {
+  flex: 0 1 calc(16.7% - 12.5px);
+}
```
frett, line 34: The number of columns is currently restricted to be 1 to 4 only. So you won't need the 5 & 6 column options

src/app/new-page/component/content-flow-item/content-flow-item.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  styleUrls: ['./content-flow-item.component.css']
+})
+export class ContentFlowItemNewComponent implements OnChanges {
+  // eslint-disable-next-line @angular-eslint/no-input-rename
+  @Input() item: FlowItem;
+  contents: Content[];
+  ready: boolean;
+  state: any;
+  isHidden: boolean;
+  flowWatcher: FlowWatcher;
```
frett, line 21: I would name this `isHiddenWatcher`, that way if in the future you start watching an additional data stream you can track the multiple active data streams and cancel all of them when appropriate

src/app/new-page/component/content-flow-item/content-flow-item.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+  }
+
+  ngOnDestroy(): void {
+    this.flowWatcher.close();
+  }
+
+  private init(): void {
+    this.isHidden = this.item.isGone(this.state);
```
frett, line 50: `watchIsGone` will emit the initial state when you start watching for changes. I don't know if the initial emission will trigger inline or as a subsequent event on the event loop though.

frett: so, setting `isHidden` here may not be necessary, but it doesn't actually hurt anything

src/app/new-page/component/content-flow-item/content-flow-item.component.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+      }
+    }
+  }
+
+  ngOnDestroy(): void {
+    this.flowWatcher.close();
+  }
+
+  private init(): void {
```
frett, line 49: can this `init()` method be called multiple times? I see the trigger is from `ngOnChanges`, which I'm assuming could be triggered multiple times.

if it is triggered multiple times you will probably want to cancel any previous active watchers before starting your new watcher.

The potential issue with not canceling a stale watcher is that it will continue to be called when the data model changes, and this will leak memory as well as potentially corrupt some UI if the stale watcher triggers a callback with some outdated/invalid state.

src/app/new-page/component/content-image/content-image.component.ts
```diff
@@ -57,6 +58,10 @@ export class ContentImageNewComponent implements OnChanges {
       this.imgResource =
         this.pageService.findAttachment(this.image.resource.name) || '';
     }
+
+    this.width = (this.image.width as any)?.value
+      ? `${(this.image.width as any)?.value}px`
+      : null;
```
frett, line 64: there are 2 types of width object that can be returned, a fixed width and a percentage width.

frett:
```typescript
export declare namespace org.cru.godtools.shared.tool.parser.model {
    abstract class Dimension {
        protected constructor();
    }
    namespace Dimension {
        class Percent extends org.cru.godtools.shared.tool.parser.model.Dimension {
            private constructor();
            get value(): number;
            copy(value?: number): org.cru.godtools.shared.tool.parser.model.Dimension.Percent;
            toString(): string;
            hashCode(): number;
            equals(other: Nullable<any>): boolean;
        }
        class Pixels extends org.cru.godtools.shared.tool.parser.model.Dimension {
            private constructor();
            get value(): number;
            copy(value?: number): org.cru.godtools.shared.tool.parser.model.Dimension.Pixels;
            toString(): string;
            hashCode(): number;
            equals(other: Nullable<any>): boolean;
        }
    }
}
```

frett: `Percent` returns a float between `0` and `1`

src/app/new-page/component/content-flow/content-flow.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+  }
+
+  trackByFn(index, item) {
+    return index;
+  }
+
+  private init(): void {
+    this.items = this.flow.items;
```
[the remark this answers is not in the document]
dr-bizz: I think I have a work around now with the code I'm about to push up.

src/app/new-page/component/content-image/content-image.component.ts
dr-bizz: I created a Dimension Parser for understanding If it's a percentage or pixel.

src/app/new-page/component/content-flow-item/content-flow-item.component.ts
dr-bizz: Gotcha, I've amened the code to ensure we close out watchers before setting up a new one.

src/app/services/xml-parser-service/xmp-parser.service.ts
```diff
@@ -117,14 +121,50 @@ export const ContentParser = (content: any): string => {
...
+  let type = null,
+    symbol = null,
+    value = null;
+  if (
+    dimension instanceof
+    org.cru.godtools.shared.tool.parser.model.Dimension.Percent
+  ) {
+    type = 'percent';
+    symbol = '%';
+    value = dimension.value;
```
frett, line 152: you'll probably want to do `value = dimension.value * 100` since you will want to present `0.5` as `50%`

src/app/new-page/component/content-flow-item/content-flow-item.component.ts
dr-bizz: When `isGone` is called, it immediately updates `this.isHidden`. While `watchIsGone` has some delay before updating `this.isHidden` with the new value.

I understand `watchIsGone` is watching to see if there are changes to the item's `isGone` value, but I'm confused about how one can change the `isGone` value to trigger the `watchIsGone`.

frett: k, the delay is from it being called as as subsequent event on the event loop, so the initial setting of `isHidden` is appropriate.

How the state can be updated is actually a bit complex, simple answer is that visibility for flow items in the emoji survey change based on which items you select on the multiselect. In this case the multiselect & flow are on different screens, so the state doesn't update "live" because it's off screen. But if the multiselect and the flow were on the same screen, then toggling an option should live update the flow items.

I believe there is a test for this live updating somewhere in renderer testing.

dr-bizz: That makes sense. Thank you for explaining. I was able to test it on the renderer testing and can confirm it's working in realtime.

dr-bizz: I'll remove the function `isGone` and only use `watchIsGone`.

src/app/new-page/component/content-card/content-card.component.html
```diff
@@ -0,0 +1,26 @@
...
+    <app-content-new-repeater [items]="contents"></app-content-new-repeater>
+  </a>
+  <a
+    *ngIf="isClickable && events.length"
+    class="card clickable"
+    (click)="eventClick()"
+    [ngStyle]="{ 'background-color': background }"
+  >
+    <app-content-new-repeater [items]="contents"></app-content-new-repeater>
+  </a>
```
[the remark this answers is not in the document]
frett: I'll create an example in renderer test

frett (approved):
everything looks good to me

PR #142, opened 2023-10-23
Adding Spacer support

dr-bizz:
## Description
Add support for Spacers on web. - https://jira.cru.org/secure/RapidBoard.jspa?rapidView=150&view=detail&selectedIssue=GT-1758

## Changes
- Added functionality for Spacer to be rendered, grabbing the height of the Spacer.
- Wrote tests to ensure properties are assigned correctly. 

I didn't do anything with `mode` as I assume this property is more for mobile.

To test, please see https://stage.knowgod.com/en/connectingwithgod/1
https://mobilecontentadmin-stage.cru.org/translations
Screenshot 2023-10-23 at 2 59 52 PM

### DIFF
--- a/src/app/new-page/component/content-spacer/content-spacer.component.css
+++ b/src/app/new-page/component/content-spacer/content-spacer.component.css
@@ -1,0 +1,5 @@
+.spacer {
+  display: block;
+  width: 100%;
+  min-height: 10px;
+}
--- a/src/app/new-page/component/content-spacer/content-spacer.component.css
+++ b/src/app/new-page/component/content-spacer/content-spacer.component.css
@@ -1,5 +1,4 @@
 .spacer {
   display: block;
   width: 100%;
-  min-height: 10px;
 }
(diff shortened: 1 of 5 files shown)


dr-bizz:
Once it passes it's test's I'll merge into staging

frett:
Mode determines if the spacer is a fixed height or if it auto scales to fill available space.

Height only really makes sense on fixed spacers.

The assumption is we won't render auto sized spacers on the web, I can get into more details tomorrow afternoon

src/app/new-page/component/content-spacer/content-spacer.component.css
```diff
@@ -0,0 +1,5 @@
+.spacer {
+  display: block;
+  width: 100%;
+  min-height: 10px;
```
frett, line 4: let's not do a `min-height`, we will depend on the content admins to define exactly the size they want

src/app/new-page/component/content-spacer/content-spacer.component.html
```diff
@@ -1 +1,6 @@
 <!-- Content Spacer Goes Here -->
+<div
+  class="spacer"
+  *ngIf="mode !== 'AUTO'"
```
frett, line 4: can we change this to be `mode === 'FIXED'`?

I'd rather do that in case we ever come up with a different spacer mode in the future

PR #149, opened 2023-11-13
GT-2180 change sorting of language dropdown menu on homepage

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=150&view=detail&selectedIssue=GT-2180
Currently the language menu on the homepage of knowgod.com sorts across the columns going left -> right and then down to the next row.
We'd like to match the way the language menu appears while you're in a tool so that it sorts going down a column, then down the next column, etc.

## Changes I made
- Changed flexbox divs to a ul so that it would sort down the column before rows.
- The number of columns changes based on screen size
- Changed CSS to make li items look the same as the divs

### DIFF
--- a/src/app/header/header.component.css
+++ b/src/app/header/header.component.css
@@ -28,7 +28,8 @@
   flex-direction: row;
   height: 100%;
   align-items: center;
-  justify-content: center;  
+  justify-content: center;
+  cursor: pointer;
 }
 
 #homeHeader > .bs-container > .language-switcher > div.selector > img {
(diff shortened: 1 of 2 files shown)


dr-bizz (commented):
I've added one comment about making the modal more center-aligned on larger screens. I know that's not something you did, but it's a easy fix, so wondered if you could throw that into this PR. I added the styles needed on a comment.

dr-bizz (approved):
Looks great!

PR #153, opened 2024-01-23
GT-2229 Fix Prayer Pathway

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=150&view=detail&selectedIssue=GT-2229#
No way to go back from button selection (ex. "Yes") without just reloading the page
Resource button does not do anything (languages other than English & Spanish)
These both may involve fixing or changing how web deals with listeners

### DIFF
--- a/src/app/page/component/tract-page/tract-page.component.ts
+++ b/src/app/page/component/tract-page/tract-page.component.ts
@@ -129,24 +129,6 @@ export class TractPageComponent implements OnChanges, OnDestroy {
       }, 0);
       return;
     } else {
-      if (functionName.includes('-no-thanks')) {
-        let isFirstPage = false;
-        let isLastPage = true;
-        this.isFirstPage$.subscribe((value) => {
-          isFirstPage = value;
-        });
-        this.isLastPage$.subscribe((value) => {
-          isLastPage = value;
-        });
-
-        if (!isLastPage) {
-          this.pageService.nextPage();
-        } else if (!isFirstPage) {
-          this.pageService.previousPage();
-        }
-        return;
-      }
-
       if (this.cards.length) {
         const cardListener = this.cards.find((card) =>
           card.listeners
@@ -176,8 +158,6 @@ export class TractPageComponent implements OnChanges, OnDestroy {
         isHideModal = !!dismissListeners.filter(
           (dismissListener) => dismissListener.name === functionName
         )?.length;
-        listeners.forEach((l) => {});
-        dismissListeners.forEach((l) => {});
       }
     }
 
@@ -214,7 +194,6 @@ export class TractPageComponent implements OnChanges, OnDestroy {
           });
       }
     } else if (isHideCard) {
-      this.next();
       setTimeout(() => {
         this.pageService.formHidden();
         this.pageService.modalHidden();
@@ -230,7 +209,6 @@ export class TractPageComponent implements OnChanges, OnDestroy {
     } else if (isHideModal) {
       this.pageService.modalHidden();
       this.pageService.formHidden();
-      this.next();
       setTimeout(() => {
         if (this._cardsHiddenOnFormAction.length) this.setHiddenCardToShow();
         if (this._cardShownOnFormAction >= 0) this.setShownCardToHidden();


frett:
I do see you removed the specific no thanks workaround. It doesn't look like you are observing page listeners yet though.

In french KGP, when you click the "Not Ready" button, one of the buttons on the displayed card is for the resources page.
Screenshot 2024-01-23 at 3 46 39 PM

When you click this button it does nothing. that button triggers the `events="resource-page-listener"` event, and the final page in KGP has the correct listener defined `listeners="resource-page-listener"`. but the web version is not jumping to that page

dr-bizz:
@frett I've created a new jira task for the page listeners error (https://jira.cru.org/browse/GT-2262), we can open a new PR to fix the page listeners

dr-bizz (commented):
From a coding perspective, this is looking good. We will need to update the production mobile content admin to add the `dismiss listeners` before we push this PR. That way we can test it on staging.

dr-bizz:
It seems the mobile app also has the same issue where you click "No thanks" and it sends you to the next page. I'm assuming this isn't correct. Hopefully adding the `dismiss listeners` will fix that for the mobile. Or it could cause issues for Mobile. We need to talk with the God Tools mobile team to check
CC @frett

frett:
On android it goes to the next page because there is also a page listener defined on that page

frett:
So removing the no thanks hard coded path on the web should be coupled with correctly supporting page listeners

frett:
here are the various events and listeners from the followup form:

followup card listeners and buttons:
```xml
    <card hidden="true" listeners="followup-form" dismiss-listeners="send-information-modal-close">

    <content:button type="event" events="followup:send send-information-modal send-information-modal-close">
    <content:button type="event" events="followup-form-no-thanks" style="outlined">
```

modal listeners and button:
```xml
    <modal listeners="send-information-modal" dismiss-listeners="done-information-modal-close">

    <content:button type="event" events="done-information-modal-close">
```

next page listeners:
```xml
    <page listeners="done-information-modal-close followup-form-no-thanks">
```

frett:
so right now based on the events & listeners, the no thanks button doesn't dismiss the followup form card, this is something that the content team can add if they desire that functionality. But it does trigger the page listener on the next page.

Submitting the form does dismiss the card and shows the modal, then clicking the button in the modal will dismiss the modal and trigger the listener on the next page.

dr-bizz (changes requested):
I believe this is being postponed, so I'm removing myself from being a reviewer.

We need to add the page listeners functionality onto God Tool before deploying this, which will require some dev hours.

src/app/page/page.component.ts
```diff
@@ -593,6 +595,69 @@ export class PageComponent implements OnInit, OnDestroy {
       .subscribe(() => {
         this.onPreviousPage();
       });
+
+    // Go to any page on page event
+    this.awaitNavigationToPageOnPageEvent();
```
dr-bizz, line 600: Maybe we call this `awaitPageEvent`

src/app/page/page.component.ts
```diff
@@ -593,6 +595,69 @@ export class PageComponent implements OnInit, OnDestroy {
...
+  private awaitNavigationToPageOnPageEvent(): void {
+    // We listen for page events and dismiss events
+    // We then navigate to the page that has the event
+    const pageListeners = this._pageBookSubPages.map((page) =>
+      page.listeners.map((listener) => listener.name)
+    );
+    const pageDismissListeners = this._pageBookSubPages.map((page) =>
+      page.dismissListeners.map((listener) => listener.name)
+    );
+    const allListenersOnAllPages = [...pageListeners, ...pageDismissListeners];
```
dr-bizz, line 612:
```suggestion
    const allListenersOnAllPages = this._pageBookSubPages.flatMap((page) => [
      ...page.listeners.map((listener) => listener.name),
      ...page.dismissListeners.map((listener) => listener.name)
    ]);
```

src/app/page/page.component.ts
```diff
@@ -593,6 +595,69 @@ export class PageComponent implements OnInit, OnDestroy {
...
+        (listener) => listener.name === event
+      );
+
+      if (foundListener || foundDismissListener) {
+        pageToNavigateTo = page;
+        isListener = foundListener;
+        isDismissListener = foundDismissListener;
+        return true;
+      }
+      return false;
```
dr-bizz, line 646: No need to return
```suggestion
      if (foundListener || foundDismissListener) {
        pageToNavigateTo = page;
        isListener = foundListener;
        isDismissListener = foundDismissListener;
      }
```

src/app/services/xml-parser-service/xmp-parser.service.ts
```diff
@@ -17,7 +17,12 @@ export type Modal = org.cru.godtools.shared.tool.parser.model.tract.Modal;
 export type TractPage =
   org.cru.godtools.shared.tool.parser.model.tract.TractPage;
 export type TractPageCard =
-  org.cru.godtools.shared.tool.parser.model.tract.TractPage.Card;
+  org.cru.godtools.shared.tool.parser.model.tract.TractPage.Card & {
+    // isHidden is a default property in the Card class that can't be edited
+    // isTemporarilyHidden is a new property so we can edit it without
+    // type errors and remembering the default isHidden value
```
dr-bizz, line 23:
```suggestion
    // type casting and this way allows us to store the default isHidden value
```

src/app/page/component/tract-page/tract-page.component.ts
```diff
@@ -121,102 +121,82 @@ export class TractPageComponent implements OnChanges, OnDestroy {
...
-            if (!card.isHidden && listener.name !== functionName) {
-              this._cardsHiddenOnFormAction.push(card.position);
-            }
-          });
-        } else {
+        const shouldHideCard =
+          !card.listeners?.some(
+            (listener) =>
+              !card.isTemporarilyHidden && listener.name === functionName
+          ) || !card.listeners?.length;
```
dr-bizz, line 184: Look into this more. Might need more explanation as it's hard to read

src/app/page/component/card/card.component.ts
```diff
@@ -20,6 +20,7 @@ export class CardComponent implements OnChanges {
   labelText: string;
   cardPosition: number;
   content: Array<Content>;
+  isHidden: boolean;
```
dr-bizz, line 23: Cane remove this

src/app/page/component/card/card.component.ts
```diff
@@ -65,6 +66,7 @@ export class CardComponent implements OnChanges {
       this.labelText = this.card.label.text?.trim() || '';
     }
 
+    this.isHidden = this.card.isTemporarilyHidden ?? this.card.isHidden;
```
dr-bizz, line 69: Can remove this

src/app/_tests/mocks.ts
```diff
@@ -405,20 +407,17 @@ export const mockTractPage = (
...
     _cards: null,
     visibleCards: null,
     backgroundImage: null,
     getAnalyticsEvents: null,
-    _dismissListeners: null,
-    _listeners: null,
-    dismissListeners: null,
-    listeners: null,
+    _dismissListeners: dismissListeners,
+    _listeners: listeners,
```
frett, line 418: I think you want to leave these as null for the mocks. they are internal data structures that I couldn't hide from the typescript headers, it was causing a compile/test error when I fully hid them previously. The actual listeners should be always be exposed via `dismissListeners` and `listeners`

I believe setting these to null will hopefully fix the current test errors, I'm also re-examining if I can truly hide them now that we are on a newer version of Kotlin

src/app/page/page.component.ts
```diff
@@ -593,6 +595,69 @@ export class PageComponent implements OnInit, OnDestroy {
...
+      this.router.navigate([
+        this._pageParams.langid,
+        this._pageParams.bookid,
+        pageToNavigateTo.position
+      ]);
+    }
+
+    if (isDismissListener) {
+      // We want to hide the page, for now I've set this to go to the next page.
+      this.onNextPage();
```
frett, line 659: hmm, I wonder if we actually do have an dismiss listeners defined for any pages (outside of CYOA where it would navigate to the previous page that launched the page).

dr-bizz:
So wierd it's failing on CI. I'll investigate

src/app/page/page.component.ts
dr-bizz: I doubt we would have dismiss listeners outside of CYOA, as they could just call go to the previous or next page.
I'll update it to go to the previous page.

dr-bizz: Confirmed with @frett that with CYOA, if page 4 navigated to page 20, the dismiss listener should navigate the user back to page 4.
---
I'll leave it to go to the next page. as CYOA isn't in this PR, and we can build the logic for that in a later PR

frett (approved):

PR #156, opened 2024-09-13
GT-2440 - Change Prev/Next Buttons to A tags

dr-bizz:
## Description
When embedded, Google doesn't pick up button clicks, so in this PR, I change the buttons that lead to a page change to A tags with a href="javascript:void(0)" attribute.


caleballdrin (approved):
Looks good

PR #157, opened 2024-10-14
Fixing Deploy GHA pipeline

dr-bizz:
## Description
In this PR, I fix the GitHub Action Deploy pipeline to prevent it from getting an error when installing AWCLI.

It seems that since our last deployment, GitHub has cracked down on the use of pip when installing an externally managed environment.

If you now want to install a global package, you need to do so in a virtual environment. To do this, we use pipe.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -60,8 +60,14 @@ jobs:
       - name: Repository checkout
         uses: actions/checkout@v4
 
-      - name: Install AWSCLI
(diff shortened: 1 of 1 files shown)


canac (approved):

frett:
it looks like the AWS CLI comes pre-installed in the GHA runner, so we may not need to manually install it like this at all.

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools

PR #158, opened 2024-10-15
GitHub Actions - AWSCLI is pre-installed so we don't need to install it

dr-bizz:
As AWSCLI is pre-installed onto the ubuntu runner, so we don't need to install it on the Github action.

I added the installation after the pipeline stopped working since GHA didn't like using `pip` to install external environments. - https://github.com/CruGlobal/know-god-web/pull/157


frett (approved):

dr-bizz:
Merged into staging and testing GHA

PR #160, opened 2024-11-01
Update to Address Security Vulnerabilities and Framework Enhancements

dr-bizz:
## Description
This PR addresses multiple security issues flagged by npm and Dependabot, involving major version upgrades and code adjustments.

### Major Fixes
- **Angular**: Upgraded from v13 to v17 to maintain security and compatibility.
- **TypeScript**: Updated from v4 to v5.2.2 to align with Angular 17 requirements.
- **core-js**: Migrated from v2 to v3, updating all references to use modern es modules instead of deprecated es6 paths.

### Key Adjustments:
**Polyfills**: Updated polyfills.ts to support core-js@3 with es module imports, ensuring compatibility with newer Angular configurations.
**Testing Enhancements**: Resolved issues in test.ts related to outdated configurations, enabling Karma to find all test files without manual imports.
**Command Optimization**: Added --watch=false to ng test due to recurring errors in watch mode. [Other users also had this error](https://github.com/angular/angular-cli/issues/28271), I followed their instructions to update packages, but after that didn't work, I ended up with this fix.
**IE11 Support Removal:** Removed support for Internet Explorer 11, as it is deprecated and unsupported by core-js@3.

### Minor Changes:
**Updated Angular.json Properties**: Refactored to use updated Angular properties for improved build configurations.
**TypeScript Target**: Set target to ES2022 for enhanced compatibility with modern JavaScript standards.


dr-bizz:
Look like I need to update the node version to node 18

.github/workflows/node.js.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Node.js setup
         uses: actions/setup-node@v4
         with:
-          node-version: '16.x'
+          node-version: '18'
```
frett, line 31: is it possible to update this to use:
```
node-version-file: .tool-versions
```

that way we don't need to update this in future upgrades?

dr-bizz: That is a good move!

frett:
@dr-bizz any reason that there is both a `package-lock.json` and `yarn.lock` file? I'd assume we probably only want to use one package manager and not both.

But my js knowledge is limited, so there very well could be a good reason I don't know

dr-bizz:
> @dr-bizz any reason that there is both a `package-lock.json` and `yarn.lock` file? I'd assume we probably only want to use one package manager and not both.
> 
> But my js knowledge is limited, so there very well could be a good reason I don't know

We can remove the `yarn.lock`, I added that by mistake.

frett:
@dr-bizz I think it was the other way, we had yarn.lock before and the package-lock.json is the new one :)

dr-bizz:
Sorry, I just saw that.

caleballdrin (approved):

PR #170, opened 2024-11-12
Security Updates - Removing protractor e2e testing

dr-bizz:
## Description
I have removed `protractor` e2e testing as it is outdated and we no longer use e2e. The baseUrl was set to a heroku domain which no longer exists.

Protractor was the reason we have a few vulnerable packages since it is no longer maintained. - https://www.npmjs.com/package/protractor

https://github.com/CruGlobal/know-god-web/security/dependabot/63
https://github.com/CruGlobal/know-god-web/security/dependabot/54
https://github.com/CruGlobal/know-god-web/security/dependabot/52


caleballdrin (approved):
Looks good

PR #171, opened 2024-11-14
Text align to use value from admin

dr-bizz:
## Description
Adding textAlign and textColor to ngStyles on the text component. Then simplified how we add the styles.

This will help with styles as the admin can control the alignment and colour from the admin tool.

### DIFF
--- a/src/app/_tests/mocks.ts
+++ b/src/app/_tests/mocks.ts
@@ -60,10 +60,13 @@ const standardTypeValues = () => {
 const createText = (text: string): Text => {
   return {
     text: text,
-    textAlign: null,
-    textColor: null,
+    textAlign: {
+      name: 'START',
+      ordinal: 0
+    },
+    textColor: '#000000',
     textScale: null,
-    _textStyles: null,
+    _textStyles: [],
     minimumLines: null,
     startImage: null,
     startImageSize: null,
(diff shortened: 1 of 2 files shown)


src/app/_tests/mocks.ts
```diff
@@ -60,10 +60,13 @@ const standardTypeValues = () => {
 const createText = (text: string): Text => {
```
frett, line 60: not something that you have to do here, just letting you know that we do expose `createTest*()` methods for several of the model objects, and the ones that are missing should be easy to expose/add.

They do have defined default values for all the parameters in Kotlin, but that may not be exported correctly to TypeScript.

```typescript
class Text extends org.cru.godtools.shared.tool.parser.model.Content {
    static createTestText(parent?: any/* org.cru.godtools.shared.tool.parser.model.Base */, text?: string, ...
}
```

caleballdrin (approved):

src/app/_tests/mocks.ts
dr-bizz: It seems like this function doesn't work properly on TS, but maybe I'm doing something wrong.

```
org.cru.godtools.shared.tool.parser.model.Text.createTestText({
    text: text
})
 ```

Returns the below object. it registers the values I input, but then doesn't set the text to those values.

```
{
  "u39_1": {
    "text": "Have you heard of the four spiritual laws?",
  },
  ...
  "text": "",
  ...
}
```

PR #172, opened 2024-11-19
Text color - No longer using color from the admin tool

dr-bizz:
Do not use color for now since we don't want to support desktop and mobile colors.


aaronlaib (approved):
Good to go!

PR #224, opened 2025-03-14
Making Github CI tests separate so it's easier to see what test failed.

dr-bizz:
## Description
In this PR, I've separated the tests into their own tests to make it easier when viewing a PR which test failed.

I've also made this change to the Give site this week, and this change would mean both repo's CI tests look alike.

Once this is approved, I'll update the branch requirements.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,4 +1,4 @@
-name: Build & Deploy
+name: CI
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


dr-bizz:
Idea -> Maybe the next PR should be to add StyleLint to God Tools

.github/workflows/node.js.yml
```diff
@@ -31,22 +30,53 @@ jobs:
...
+
+      - name: 📦 Install dependencies
+        run: yarn install --immutable --immutable-cache
+
       - name: 🏗️ Build app
         run: yarn build
 
   deploy:
-    name: Deploy
+    name: 🚀 Deploy app
```
frett, line 79: do we want to make Deploy dependent on all the other jobs, or just on `build`?

dr-bizz: Good call

frett (approved):

PR #229, opened 2025-03-20
GT-2527 - Implement Flows attributes and styling

caleballdrin:
GT-2527

- Implement Flow `item-width` and `row-gravity` using flexbox CSS styles
- Implement Paragraph `text-scale`, `minimum-lines` and `start-image`

### DIFF
--- a/src/app/page/component/content-flow-item/content-flow-item.component.css
+++ b/src/app/page/component/content-flow-item/content-flow-item.component.css
@@ -1,11 +1,28 @@
 .flow-item {
-  width: 130px;
-  margin-left: 7.5px;
-  margin-right: 7.5px;
-  padding: 5px;
   margin-top: 0px;
+  height: 100%;
+  flex-grow: 0;
+  flex-shrink: 0;
+  box-sizing: border-box;
+}
+
+.flow-item app-content-repeater app-content-card > a {
+  height: 100%;
 }
 
 .flow-item app-content-repeater {
+  display: block;
+  height: 100%;
   margin-top: 0;
+}
+
+.flow-item p.textContent {
+  margin-top: 0 !important;
+  align-items: center;
+  width: 100%;
+  margin-left: auto;
+  margin-right: auto;
+}
+.flow-item app-content-text {
+  margin-top: 0 !important;
 }
--- a/src/app/page/component/content-flow-item/content-flow-item.component.ts
+++ b/src/app/page/component/content-flow-item/content-flow-item.component.ts
@@ -3,7 +3,8 @@ import {
   Input,
   OnChanges,
   OnDestroy,
-  SimpleChanges
+  SimpleChanges,
+  ViewEncapsulation
 } from '@angular/core';
 import {
   Content,
@@ -15,7 +16,8 @@ import { PageService } from '../../service/page-service.service';
 @Component({
   selector: 'app-content-flow-item',
   templateUrl: './content-flow-item.component.html',
-  styleUrls: ['./content-flow-item.component.css']
+  styleUrls: ['./content-flow-item.component.css'],
+  encapsulation: ViewEncapsulation.None
 })
 export class ContentFlowItemComponent implements OnChanges, OnDestroy {
   @Input() item: FlowItem;
--- a/src/app/page/component/content-flow/content-flow.component.html
+++ b/src/app/page/component/content-flow/content-flow.component.html
@@ -5,11 +5,12 @@
       justifyContent: justifyContent
     }"
   >
+    <!-- Subtract 17.5px from the percentage to account for the gap between flow-items  -->
     <app-content-flow-item
       *ngFor="let item of items; trackBy: trackByFn"
       [item]="item"
       [ngStyle]="{
-        flexBasis: 'calc(' + item.itemWidth + ' - 20px)'
+        flexBasis: 'calc(' + item.itemWidth + ' - 17.5px)'
       }"
     >
     </app-content-flow-item>
--- a/src/app/page/component/content-flow/content-flow.component.html
+++ b/src/app/page/component/content-flow/content-flow.component.html
@@ -1,8 +1,16 @@
 <ng-container *ngIf="ready">
-  <div class="flow mw568">
+  <div
+    class="flow mw568"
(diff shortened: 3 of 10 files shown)


frett (approved):
from what I can tell this looks good

src/app/page/component/content-flow/content-flow.component.html
```diff
@@ -1,8 +1,16 @@
...
+    [ngStyle]="{
+      justifyContent: justifyContent
+    }"
+  >
     <app-content-new-flow-item
       *ngFor="let item of items; trackBy: trackByFn"
       [item]="item"
+      [ngStyle]="{
+        flexBasis: 'calc(' + item.itemWidth + ' - 20px)'
+      }"
```
dr-bizz, line 13: Why minus 20px?

src/app/page/component/content-flow-item/content-flow-item.component.ts
```diff
@@ -15,7 +16,8 @@ import { PageService } from '../../service/page-service.service';
 @Component({
   selector: 'app-content-new-flow-item',
   templateUrl: './content-flow-item.component.html',
-  styleUrls: ['./content-flow-item.component.css']
+  styleUrls: ['./content-flow-item.component.css'],
+  encapsulation: ViewEncapsulation.None
```
dr-bizz, line 20: Do you still need this as all your styles are for this component?

src/app/page/component/content-flow-item/content-flow-item.component.css
```diff
@@ -1,12 +1,28 @@
...
 .flow-item app-content-new-repeater {
+  display: block;
+  height: 100%;
   margin-top: 0;
-}
\ No newline at end of file
+}
+
+.flow-item p.textContent {
+  margin-top: 0 !important;
```
dr-bizz, line 20: I assume that you wouldn't have added `!important` unless you truly needed it, but just in case, we only should add this as a last resort.

dr-bizz (approved):
Great job on this! It's looking great.

I know we have a mobile app for this so most user's won't see this but on smaller screens, the flow items aren't centered due to the gap on the right side.

I'm going to approve it as I don't want to block you since I'm out next week.

src/app/page/component/content-flow-item/content-flow-item.component.ts
caleballdrin: I need this because I need some of the `content-flow-item` styles to apply to the parent components.

src/app/page/component/content-flow-item/content-flow-item.component.css
caleballdrin: Yes, I need it to override the general style of paragraphs, which always has a top margin (when not the first child). The Openers tool uses the paragraph, and we don't want it to apply the margin.

frett: Overall I would be careful with applying one off styling overrides. In reality the content should be styled consistent based upon what is in the content xml.

A paragraph element in the content xml should be rendered with a small amount of top and bottom padding.

package.json
```diff
@@ -25,7 +25,7 @@
     "@angular/platform-browser": "^17.3.12",
     "@angular/platform-browser-dynamic": "^17.3.12",
     "@angular/router": "^17.3.12",
-    "@cruglobal/godtools-shared": "^1.1.2-SNAPSHOT.1686",
+    "@cruglobal/godtools-shared": "^1.1.3-SNAPSHOT.1705",
```
frett, line 28: I would prefer this gets updated to the stable release of 1.1.2 instead of the snapshot build since font-weight will be in the 1.1.2 stable build.

The snapshot builds may be unstable since it's an automated build based off the current development branch. It's something that's useful for testing out new functionality, but when it's not necessary I'd like to stay on the stable releases that get tagged `latest` in npm.

Screenshot 2025-04-08 at 1 41 30 PM

dr-bizz: We can use version 1.1.2. But I believe we started to use the snapshot as it had a feature which wasn't released yet. It might be released on the latest stable version now.

But I understand your concern about building against the stable versions as there could be errors with the snapshots.

dr-bizz: @caleballdrin was there a reason you upgraded it to `1.1.3-SNAPSHOT.1705` instead of `1.1.2`?

src/app/page/component/content-flow-item/content-flow-item.component.css
dr-bizz: One of the issues we have encountered previously with the styles arises from the fact that the web design is different from the mobile design, necessitating the need to disregard certain styles.

I doubt that this is occurring here; rather, it seems to be related to the XML styling the margin, along with the web app having its own styles that are conflicting.

dr-bizz (approved):
Great work on this. I've left some comments, but not much

package.json
caleballdrin: Yeah I was just needing to access things from the latest snapshot. But now I have it back to 1.1.2.

src/app/page/component/content-text/content-text.component.ts
```diff
@@ -51,11 +51,11 @@ export class ContentTextComponent implements OnChanges {
...
   private init(): void {
     const styles = {
-      'font-weight': this.text.textStyles?.some(
-        (style) => style.name === 'BOLD'
-      )
-        ? 'bold'
-        : '',
+      'font-weight': this.text.fontWeight
+        ? this.text.fontWeight
+        : this.text.textStyles?.some((style) => style.name === 'BOLD')
```
frett, line 56: You no longer need to check for a `BOLD` textStyle, when I added fontWeight I deprecated the bold style and handled automatic fallback within the shared library for you.

frett: relevant fallback logic: https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/blob/main/module/parser/src/commonMain/kotlin/org/cru/godtools/shared/tool/parser/model/Text.kt#L95-L96

caleballdrin: Great, thank you!

PR #230, opened 2025-03-24
Upgrade to node 20

dr-bizz:
## Description
In this PR, I've upgraded Node to version `20.17.0`; it was on version 18 before. I also upgraded Yarn to version 4, as our other repos use Yarn 4 with Node 20.
I could have upgraded Node to `20.18.3`, but since the `@types/node` version was released last week and no `@types/node` version is available for that release, I opted for `20.17.0`.

Jira: [GT-2539](https://jira.cru.org/browse/GT-2539)

### Changes
- I've updated the README file to include the new node version.
- Updated packages which needed to be updated.
- Upgraded Yarn to version 4.

## Testing:
I'll push this to development and staging

## Checklist:

- [ ] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [ ] I have applied the appropriate labels. (_Add the label "On Staging" to get the branch automatically merged into staging._)
- [ ] I have requested a review from another person on the project


canac:
@dr-bizz Just ping me or request another review when you're ready

frett:
@dr-bizz now that yarn 4 adds the binary zip files for dependencies to the repo, would it make sense to use git lfs for those binary files?

we currently use LFS for our reference screenshots for GodTools screenshot tests, it keeps large binary files out of the the git repo

we can talk about pros/cons on slack if you want

canac:
> would it make sense to use git lfs for those binary files?

Ah, we should have done this from the beginning for MPDX. I assume since they're in the git history, the current versions will be there forever even if we switch to Git LFS.

frett:
> > would it make sense to use git lfs for those binary files?
> 
> Ah, we should have done this from the beginning for MPDX. I assume since they're in the git history, the current versions will be there forever even if we switch to Git LFS.

yeah, unless you rewrite history to remove it, but that leads to more problems than it's worth.

dr-bizz:
I'll clean up the Commit history once reviews are complete

frett:
I'll leave it up to you web guys to sort out if we want to track the yarn cache or not. You are more familiar with the benefits of it.

If the cache remains in git, I think it makes sense to continue to use LFS for the files. if we remove it, then there is no need for git lfs

.github/workflows/node.js.yml
```diff
@@ -45,13 +48,16 @@ jobs:
...
+        run: corepack enable
+
       - name: Node.js setup
         uses: actions/setup-node@v4
         with:
           node-version-file: .tool-versions
 
       - name: 📦 Install dependencies
-        run: yarn install --immutable --immutable-cache
+        run: yarn install
```
canac, line 60: I was going to recommend keeping `--immutable`, but `yarn` v4 is smart and defaults `--immutable` to true in CI.

canac (commented):
This looks great! It builds, tests, and runs great locally.

.github/workflows/node.js.yml
dr-bizz: Is there something you are wanting me to do before you approve this?

canac (approved):
Sorry, I meant to click approve

dr-bizz:
I just added a Jira ticket to the title, but it already had it in the description. - [GT-2539](https://jira.cru.org/browse/GT-2539)

PR #244, opened 2025-04-14
Removing 'new' from component names as it's no longer needed.

dr-bizz:
## Description
In this PR, I have renamed the components so we no longer have `new` in the name.

I added `new` in the beginning because we had to build the new web page alongside the old one. Now that the old one is removed, we can rename all the component names.


dr-bizz:
I created it from the Navigation PR. Once approved I'll redo the commit history to make only 1 commit (Updating the names of the components)

PR #245, opened 2025-04-15
GT-2528 CYOA Card Collection

caleballdrin:
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=150&projectKey=GT&view=detail&selectedIssue=GT-2528#

**Add CYOA Card Collection Page**

- Navigation arrows
- Pass the total cards to each card

**Add CYOA Card Component**

- Show Card position and total cards

**Make the URL include cardPosition**

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -29,7 +29,9 @@ import { ContentSpacerComponent } from './page/component/content-spacer/content-
 import { ContentTabsComponent } from './page/component/content-tabs/content-tabs.component';
 import { ContentTextComponent } from './page/component/content-text/content-text.component';
 import { ContentVideoComponent } from './page/component/content-video/content-video.component';
+import { CyoaCardComponent } from './page/component/cyoa/card/card.component';
 import { ModalComponent } from './page/component/modal/modal.component';
+import { CYOACardCollectionComponent } from './page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component';
 import { CYOAComponent } from './page/component/page/cyoa-page/cyoa-page.component';
 import { TractPageComponent } from './page/component/page/tract-page/tract-page.component';
 import { PageHeaderComponent } from './page/component/page-header/page-header.component';
@@ -47,6 +49,7 @@ import { SharingModalComponent } from './shared/sharing-modal/sharing-modal.comp
 // :resourceType - The resource type 'v1' for tract or 'v2' for cyoa
 // :bookId - The book ID ('thefour', 'kgp', etc.)
 // :page - The page number
+// :cardPosition - The card position in the card collection
 //============
 
 const appRoutes: Routes = [
@@ -69,6 +72,10 @@ const appRoutes: Routes = [
   },
   {
     path: ':langId/:toolType/:resourceType/:bookId/:page',
+    component: PageComponent
+  },
+  {
+    path: ':langId/:toolType/:resourceType/:bookId/:page/:cardPosition',
     component: PageComponent
   }
 ];
@@ -110,7 +117,9 @@ export function playerFactory() {
     ContentFlowComponent,
     ContentFlowItemComponent,
     ContentCardComponent,
-    CYOAComponent
+    CyoaCardComponent,
+    CYOAComponent,
+    CYOACardCollectionComponent
   ],
   imports: [
     BrowserModule,
--- a/src/app/page/component/cyoa/card/card.component.html
+++ b/src/app/page/component/cyoa/card/card.component.html
@@ -1,6 +1,5 @@
 <div attr.id="card-{{ cardPosition }}" class="cardContent">
-  <h4 class="mw568 cardHeading ng-binding" dir="{{ dir$ }}"></h4>
-  <ng-container *ngIf="content && (isModal$ | async) === false">
+  <ng-container *ngIf="content">
     <app-content-new-repeater [items]="content"></app-content-new-repeater>
     <p class="cardPosition">{{ cardPosition + 1 }} / {{ totalCards }}</p>
   </ng-container>
--- a/src/app/page/component/cyoa/card/card.component.html
+++ b/src/app/page/component/cyoa/card/card.component.html
@@ -1,0 +1,7 @@
+<div attr.id="card-{{ cardPosition }}" class="cardContent">
+  <h4 class="mw568 cardHeading ng-binding" dir="{{ dir$ }}"></h4>
+  <ng-container *ngIf="content && (isModal$ | async) === false">
+    <app-content-new-repeater [items]="content"></app-content-new-repeater>
+    <p class="cardPosition">{{ cardPosition + 1 }} / {{ totalCards }}</p>
+  </ng-container>
+</div>
--- a/src/app/page/component/cyoa/card/card.component.html
+++ b/src/app/page/component/cyoa/card/card.component.html
@@ -1,6 +1,6 @@
 <div attr.id="card-{{ cardPosition }}" class="cardContent">
   <ng-container *ngIf="content">
-    <app-content-new-repeater [items]="content"></app-content-new-repeater>
+    <app-content-repeater [items]="content"></app-content-repeater>
     <p class="cardPosition">{{ cardPosition + 1 }} / {{ totalCards }}</p>
   </ng-container>
 </div>
--- a/src/app/page/component/cyoa/card/card.component.ts
+++ b/src/app/page/component/cyoa/card/card.component.ts
@@ -25,18 +25,15 @@ export class CyoaCardComponent implements OnChanges {
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    for (const propName in changes) {
-      if (changes.hasOwnProperty(propName)) {
-        if (
-          propName === 'card' &&
-          (!changes['card'].previousValue ||
-            changes['card'].currentValue !== changes['card'].previousValue)
-        ) {
-          this.ready = false;
-          this.cardPosition = 0;
-          this.content = [];
-          this.init();
-        }
+    if (changes.hasOwnProperty('card')) {
+      if (
+        !changes.card.previousValue ||
+        changes.card.currentValue !== changes.card.previousValue
+      ) {
+        this.ready = false;
+        this.cardPosition = 0;
+        this.content = [];
+        this.init();
       }
     }
   }
--- a/src/app/page/component/cyoa/card/card.component.ts
+++ b/src/app/page/component/cyoa/card/card.component.ts
@@ -1,0 +1,66 @@
+import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Observable } from 'rxjs';
+import { PageService } from 'src/app/page/service/page-service.service';
+import {
+  CYOAPageCard,
+  Content
+} from 'src/app/services/xml-parser-service/xmp-parser.service';
+
+@Component({
+  selector: 'app-page-cyoa-card',
+  templateUrl: './card.component.html',
+  styleUrls: ['./card.component.css']
+})
+export class CyoaCardComponent implements OnChanges {
+  @Input() card: CYOAPageCard;
+  @Input() totalCards: number;
+
+  ready: boolean;
+  cardPosition: number;
+  content: Array<Content>;
+  dir$: Observable<string>;
+  isForm$: Observable<boolean>;
+  isModal$: Observable<boolean>;
+  isFirstPage$: Observable<boolean>;
+
+  constructor(private pageService: PageService) {
+    this.dir$ = this.pageService.pageDir$;
+    this.isForm$ = this.pageService.isForm$;
+    this.isModal$ = this.pageService.isModal$;
+    this.isFirstPage$ = this.pageService.isFirstPage$;
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    for (const propName in changes) {
+      if (changes.hasOwnProperty(propName)) {
+        switch (propName) {
+          case 'card': {
+            if (
+              !changes['card'].previousValue ||
+              changes['card'].currentValue !== changes['card'].previousValue
+            ) {
+              this.ready = false;
+              this.cardPosition = 0;
+              this.content = [];
+              this.init();
+            }
+          }
+        }
+      }
+    }
+  }
+
+  trackByFn(index) {
+    return index;
+  }
+
+  private init(): void {
+    this.cardPosition = this.card.position || 0;
+
+    this.content = this.card.content;
+
+    console.log('card ', this.card);
+
+    this.ready = true;
+  }
+}
--- a/src/app/page/component/cyoa/card/card.component.ts
+++ b/src/app/page/component/cyoa/card/card.component.ts
@@ -7,7 +7,7 @@ import {
 } from 'src/app/services/xml-parser-service/xmp-parser.service';
 
 @Component({
-  selector: 'app-page-cyoa-card',
+  selector: 'app-cyoa-card',
   templateUrl: './card.component.html',
   styleUrls: ['./card.component.css']
 })
@@ -19,47 +19,32 @@ export class CyoaCardComponent implements OnChanges {
   cardPosition: number;
   content: Array<Content>;
   dir$: Observable<string>;
-  isForm$: Observable<boolean>;
-  isModal$: Observable<boolean>;
-  isFirstPage$: Observable<boolean>;
 
-  constructor(private pageService: PageService) {
+  constructor(private readonly pageService: PageService) {
     this.dir$ = this.pageService.pageDir$;
-    this.isForm$ = this.pageService.isForm$;
-    this.isModal$ = this.pageService.isModal$;
-    this.isFirstPage$ = this.pageService.isFirstPage$;
   }
 
   ngOnChanges(changes: SimpleChanges) {
     for (const propName in changes) {
       if (changes.hasOwnProperty(propName)) {
-        switch (propName) {
-          case 'card': {
-            if (
-              !changes['card'].previousValue ||
-              changes['card'].currentValue !== changes['card'].previousValue
-            ) {
-              this.ready = false;
-              this.cardPosition = 0;
-              this.content = [];
-              this.init();
-            }
-          }
+        if (
+          propName === 'card' &&
+          (!changes['card'].previousValue ||
+            changes['card'].currentValue !== changes['card'].previousValue)
+        ) {
+          this.ready = false;
+          this.cardPosition = 0;
+          this.content = [];
+          this.init();
         }
       }
     }
-  }
-
-  trackByFn(index) {
-    return index;
   }
 
   private init(): void {
     this.cardPosition = this.card.position || 0;
 
     this.content = this.card.content;
-
-    console.log('card ', this.card);
 
     this.ready = true;
   }
--- a/src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
+++ b/src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
@@ -1,5 +1,13 @@
 <ng-container *ngIf="ready">
   <div id="toolContent">
+    <a
+      (click)="navigateBack()"
+      href="javascript:void(0)"
+      *ngIf="showBackButton"
+      class="backwards-navigation"
+    >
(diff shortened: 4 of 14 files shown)


src/app/page/page.component.html
```diff
@@ -38,13 +38,21 @@
...
-        *ngIf="resourceType === 'cyoa'"
+        *ngIf="resourceType === 'cyoa' && !activePage.hasOwnProperty('_cards')"
         [page]="activePage"
         [order]="activePageOrder"
         [totalPages]="totalPages"
       >
       </app-cyoa-page>
 
+      <app-cyoa-card-collection-page
+        *ngIf="resourceType === 'cyoa' && activePage.hasOwnProperty('_cards')"
```
frett, line 49: is there a way to do this if based upon the type of `activePage` instead of if it has an internal property? if not would creating/exposing a `page.type` attribute be useful for this?

caleballdrin: I tried doing `activePage instanceOf CyoaCardCollectionPage,` but that didn't work. I think a `page.type` attribute would be ideal.

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+        } else {
+          this.goToCard(index);
+        }
+      });
+  }
+
+  private init(): void {
+    console.log('CYOA Card Collection Page:', this._page);
+    console.log('Order:', this.order);
+    console.log('Total Pages:', this.totalPages);
```
dr-bizz, line 139: Remove console logs
```suggestion
```

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.pageService.setPageOrder(this.order, this.totalPages);
+    this.pageService.modalHidden();
+    this.pageService.formHidden();
+    this.cards = this._page.cards;
+    this.totalCards = this.cards.length || 0;
+    this.showBackButton = !!this._page.parentPage?.position;
```
dr-bizz, line 145: I'm thinking anywhere we have repeating code we should put in a helper function

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+      relativeTo: this.route
+    });
+  }
+
+  private updateCardState(): void {
+    this.isFirstCard = this.currentCardIndex === 0;
+    this.isLastCard = this.currentCardIndex === this.totalCards - 1;
+  }
+
+  private subscribeToCardPosition(): void {
```
dr-bizz, line 118: We normally name this with `on` at the front of the function name.
For example: `onCardPositionChange`

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.ready = true;
+  }
+
+  navigateBack(): void {
+    if (!this.ready || !this.showBackButton) {
+      return;
+    }
+
+    this.pageService.navigateToPage(this._page.parentPage.position);
+  }
```
dr-bizz, line 157: Same here, we prob should put this in a helper function to help maintain this code

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+      this.pageService.emailSignumFormDataNeeded();
+      setTimeout(() => {
+        this.onFormAction(functionName);
+      }, 0);
+      return;
+    }
+
+    console.log('Function Name:', functionName);
+  }
+}
```
dr-bizz, line 179: Same here if this is the same as the cyoa-page

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+  currentCardIndex = 0;
+
+  get currentCard(): CYOAPageCard | undefined {
+    return this.cards?.[this.currentCardIndex];
+  }
+
+  goToCard(index: number): void {
+    if (!this.cards || this.cards.length === 0) return;
```
dr-bizz, line 88:
```suggestion
    if (!this.cards || this.cards.length === 0) {
      return;
    }
```

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const safeIndex = Math.max(0, Math.min(index, this.cards.length - 1));
```
dr-bizz, line 90: This is great! It might be wise to add this to the next and previous functions

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+              this.init();
+            }
+            break;
+          default:
+            break;
+        }
+      }
+    }
+  }
+  currentCardIndex = 0;
```
dr-bizz, line 81: This shouldn't be here, it should either be in a function, like the `init()` or it should be with `showBackButton` at the top

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      { id: '1', content: [] },
+      { id: '2', content: [] }
+    ] as any;
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should update the URL when showing next card', () => {
```
dr-bizz, line 49: Lets also test that the `isFirstCard` and `isLastCard` is updated

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  it('should show previous card', () => {
+    component.currentCardIndex = 1;
+    spyOn(router, 'navigate');
+
+    component.showPreviousCard();
+
+    expect(component.currentCardIndex).toBe(0);
+    expect(router.navigate).toHaveBeenCalledWith(['../', 0], {
+      relativeTo: activatedRoute
+    });
```
dr-bizz, line 90: Lets also add a test for the `subscribeToCardPosition` function

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
```diff
@@ -0,0 +1,89 @@
+<ng-container *ngIf="ready">
+  <div id="toolContent">
+    <a
+      (click)="navigateBack()"
+      href="javascript:void(0)"
+      *ngIf="showBackButton"
+      class="backwards-navigation"
+    >
+      <img src="assets/img/back-arrow.png" alt="navigate back" />
+    </a>
```
dr-bizz, line 10: I might need to create a component for this, so we don't have to repeat

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
```diff
@@ -0,0 +1,89 @@
...
+            >Learn more at godtoolsapp.com</a
+          ></span
+        >
+        <span
+          >© 1994-{{ currentYear }}
+          <a href="http://cru.org" target="_blank">Cru</a>. All Rights
+          Reserved.</span
+        >
+      </footer>
+    </ng-container>
```
dr-bizz, line 87: I wonder how much of this we could put into a component so we don't have to repeat code between 2 or 3 pages

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.css
dr-bizz, line 1: We could remove this page

src/app/page/component/cyoa/card/card.component.ts
```diff
@@ -0,0 +1,66 @@
...
+  trackByFn(index) {
+    return index;
+  }
+
+  private init(): void {
+    this.cardPosition = this.card.position || 0;
+
+    this.content = this.card.content;
+
+    console.log('card ', this.card);
```
dr-bizz, line 62:
```suggestion
```

src/app/page/component/cyoa/card/card.component.ts
```diff
@@ -0,0 +1,66 @@
...
+              !changes['card'].previousValue ||
+              changes['card'].currentValue !== changes['card'].previousValue
+            ) {
+              this.ready = false;
+              this.cardPosition = 0;
+              this.content = [];
+              this.init();
+            }
+          }
+        }
```
dr-bizz, line 48: As we're only using one clause for the switch statement, it might be better to just switch it to a If statement.
I also know we prob do this on a lot of other components.

src/app/page/component/cyoa/card/card.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  ready: boolean;
+  cardPosition: number;
+  content: Array<Content>;
+  dir$: Observable<string>;
+  isForm$: Observable<boolean>;
+  isModal$: Observable<boolean>;
+  isFirstPage$: Observable<boolean>;
+
+  constructor(private pageService: PageService) {
```
dr-bizz, line 26:
```suggestion
  constructor(private readonly pageService: PageService) {
```

src/app/page/component/cyoa/card/card.component.ts
```diff
@@ -0,0 +1,66 @@
+import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Observable } from 'rxjs';
+import { PageService } from 'src/app/page/service/page-service.service';
+import {
+  CYOAPageCard,
+  Content
+} from 'src/app/services/xml-parser-service/xmp-parser.service';
+
+@Component({
+  selector: 'app-page-cyoa-card',
```
dr-bizz, line 10: Do you need the name `page` in the selector? I don't see the name `page` anywhere else on this file

src/app/page/component/cyoa/card/card.component.html
```diff
@@ -0,0 +1,7 @@
+<div attr.id="card-{{ cardPosition }}" class="cardContent">
+  <h4 class="mw568 cardHeading ng-binding" dir="{{ dir$ }}"></h4>
```
dr-bizz, line 2: Empty H4
```suggestion
```

dr-bizz (changes requested):
Great work on this! Keep up the good work. I have left a few comments, but it's looking good.

src/app/page/page.component.html
frett: it looks like knowgod already checks the type for content elements:

https://github.com/CruGlobal/know-god-web/blob/main/src/app/services/xml-parser-service/xmp-parser.service.ts#L58-L136

I wonder if there are restrictions for using `instanceof` within an angular expression.

You could expose an additional property on the `PageComponent` that does the type checking for you and returns a constant string for the type, something like:
```
activePageType: string;

fun showPage(page: Page) {
  ...

  if (page instanceof org.cru.godtools.shared.tool.parser.model.page.ContentPage) {
    activePageType = 'content'
  } else if (page instanceof ...) {
    activePageType = ...
  } else {
    activePageType = ''
  }
}
```

src/app/app.module.ts
```diff
@@ -70,6 +73,10 @@ const appRoutes: Routes = [
   {
     path: ':langId/:toolType/:resourceType/:bookId/:page',
     component: PageComponent
+  },
+  {
+    path: ':langId/:toolType/:resourceType/:bookId/:page/:cardPosition',
+    component: PageComponent
   }
```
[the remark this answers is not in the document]
caleballdrin: Adding `cardPosition` breaks other pages that don't have `cardPositions`.

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
caleballdrin: I can do that in a seperate PR

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
caleballdrin: I can do that in a seperate PR

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
caleballdrin: The next and previous functions call this function goToCard()

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  private updateUrl(): void {
+    this.router.navigate(['../', this.currentCardIndex], {
+      relativeTo: this.route
+    });
+  }
+
+  private updateCardState(): void {
+    this.isFirstCard = this.currentCardIndex === 0;
+    this.isLastCard = this.currentCardIndex === this.totalCards - 1;
```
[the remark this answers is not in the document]
canac: For example, I broke this logic by going to http://localhost:4200/en/tool/v2/openersstage/4/foo. It seems that the page index is 0 in the UI, but the back button is shown.

canac (changes requested):
I couldn't see the page content, but that may have been an issue on my end. I have a few minor suggestions. Feel free to push back on anything because I'm not super familiar with this codebase.

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  showNextCard(): void {
+    this.goToCard(this.currentCardIndex + 1);
+  }
+
+  private updateUrl(): void {
+    this.router.navigate(['../', this.currentCardIndex], {
+      relativeTo: this.route
+    });
```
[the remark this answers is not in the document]
caleballdrin: @canac do you have any thoughts on this?

src/app/page/component/page/page-helpers.ts
```diff
@@ -0,0 +1,17 @@
+export function shouldShowBackButton(page: {
+  parentPage?: { position?: any };
```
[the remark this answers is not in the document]
caleballdrin: @dr-bizz @frett Will parent page always be referenced by it's position? I know we are going to start using the `id` in the url instead of `position`.

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
```diff
@@ -0,0 +1,156 @@
...
+      if (changes.hasOwnProperty(propName)) {
+        switch (propName) {
+          case 'page':
+            if (
+              !changes['page'].previousValue ||
+              changes['page'].currentValue !== changes['page'].previousValue
+            ) {
+              this.ready = false;
+              this._page = this.page;
+              this.cards = [];
```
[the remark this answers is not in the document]
caleballdrin: I get errors when I remove this. I think it is because `ngOnChanges` seems to be run before `init()`. So I get an error saying `this._page` is not defined.

src/app/page/component/page/page-helpers.ts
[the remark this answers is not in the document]
dr-bizz: We are switching the URL to have the name instead of the position on CYOA, but not for the original tools.

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.ts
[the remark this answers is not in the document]
canac: I'm not super concerned about using a relative route. But if that's a concern, I would pass a `setUrlCard(card: number)` (the name could be improved) function from the page component to this component. This component can call it to set the URL. That way the page component is be entirely responsible for building out the correct URL given a card index. I like that better than duplicating logic from the page component in this component.

src/app/page/page.component.ts
```diff
@@ -126,6 +129,40 @@ export class PageComponent implements OnInit, OnDestroy {
     }
   }
 
+  setPageType(activePage): string {
```
canac, line 132: This now gets the page type instead of setting it.

Also, I think we can remove the `activePage` argument and reference `this.activePage` directly in the function body.

```suggestion
  getPageType(): string {
```

src/app/page/page.component.ts
```diff
@@ -115,6 +117,7 @@ export class PageComponent implements OnInit, OnDestroy {
     this.route.queryParams.subscribe((queryParam) => {
       this.embedded = queryParam.embedded === 'true';
     });
+    this.activePageType = this.setPageType(this.activePage);
```
canac, line 120: Instead of making an instance variable, I was thinking we could call `this.getPageType()` from within the template. Is that doable?

src/app/page/page.component.ts
```diff
@@ -126,6 +129,40 @@ export class PageComponent implements OnInit, OnDestroy {
...
+      org.cru.godtools.shared.tool.parser.model.page.CardCollectionPage
+    ) {
+      return 'card-collection';
+    } else {
+      return '';
+    }
+  }
+
+  setCardUrl = (card: number) => {
+    if (!this._pageParams?.langId) {
```
canac, line 149: I think that `this._pageParams` is always an object.

```suggestion
    if (!this._pageParams.langId) {
```

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        PageService
+      ]
+    }).compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CYOACardCollectionComponent);
+    component = fixture.componentInstance;
+    component.setCardUrl = (card: number) => card;
+    setCardUrlSpy = spyOn(component, 'setCardUrl').and.callThrough();
```
canac, line 37: I think that if you remove `.and.callThrough()`, the original method isn't called, so you wouldn't have to overwrite it.

```suggestion
    setCardUrlSpy = spyOn(component, 'setCardUrl');
```

canac (commented):
Just a couple more suggestions, but this looks nearly done! Great work!

src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.spec.ts
caleballdrin: I get an error when doing that. `Error: <spyOn> : setCardUrl() method does not exist`

canac: So sorry, I didn't realize this was the other component and that method is an input.

src/app/page/component/page/page-helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+}): boolean {
+  return !!page.parentPage?.position;
```
canac, line 4: Is it OK that this will return false if `page.parentPage.position === 0`? It's slightly different from the `page.parentPage?.position !== undefined` check in `navigateBackIfPossible`. I think it would be good to use the same check unless there's a reason for them to be different.

canac (approved):
Code looks great, and tests great too! The content is loading for me now too.

I still don't love the `position: any` types in `page-helpers.ts`, but I guess that's not necessarily blocking. Even if it's going to change to a string soon, I think it would be good to set it to `number` for now.

PR #247, opened 2025-04-24
GT-2574-CYOA-urls

caleballdrin:
Use page id in CYOA pages.

### DIFF
--- a/src/app/_tests/mocks.ts
+++ b/src/app/_tests/mocks.ts
@@ -62,6 +62,7 @@ const standardTypeValues = () => {
     _events: null,
     _getAnalyticsEvents: null,
     getAnalyticsEvents: null,
+    controlColor: null,
     equals: () => null,
     hashCode: () => null
   };
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -59,15 +59,14 @@ const appRoutes: Routes = [
     redirectTo: ':langId/tool/v1/:bookId/:page',
     pathMatch: 'full'
   },
-  {
-    path: ':langId/:toolType/:resourceType/:bookId',
-    redirectTo: ':langId/:toolType/:resourceType/:bookId/0',
-    pathMatch: 'full'
-  },
   { path: ':langId', component: DashboardComponent },
   { path: '', component: DashboardComponent },
   {
     path: ':langId/embed/:toolType/:resourceType/:bookId',
+    component: PageComponent
+  },
+  {
+    path: ':langId/:toolType/:resourceType/:bookId',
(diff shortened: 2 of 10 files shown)


caleballdrin:
@dr-bizz One of the tests you added on getResources() is failing. Should that be included in this PR? It seems unrelated to URL navigation.

dr-bizz (commented):
I see we still have the issue of sending the user to `position` first and then switching to the `name`.

Were you able to locate why it does this?

The function `navigateToPage` on `dashbaord.component.ts` has the correct URL, using the name, but it still goes to the URL with the position.

dr-bizz:
@frett With the switch from using the page's ID instead of the POSITION in the new URL change, we cannot find the page name from the dashboard. This means we have to send the user to `openers/0`, which will then redirect to `intro`.

This is not the best solution since it's causing a redirect.

Would you be able to add the first page's name to the resource's attributes? This would mean we wouldn't have to load each book's pages to find the first page for each CYOA book.

![Screenshot 2025-05-07 at 11 52 43 AM](https://github.com/user-attachments/assets/84dfbc76-3226-4cb5-9638-870d444714bc

frett:
hmm, I'll think about how to accomplish it, those attributes are disconnected from the actual content.

I know this doesn't address the extra redirect, but would going to `/openers` without a page index and allowing that to update the URL to `/openers/intro` once the content xml is loaded work? we are already intending to have that redirect work as an easy way to start a tool at the beginning

caleballdrin:
@dr-bizz it sounds like Frett is suggesting what we currently have. In the past, it has always redirected to `position` 0, now it just also changes the URL to the `id`.

dr-bizz:
So it is currently redirecting 2 times. First from `/openers` to `/openers/0`, and then from `/openers/0` to `/openers/intro`?


The code in `src/app/app.module.ts` does the redirect from `/openers` to `/openers/0`

```
{
    path: ':langId/:toolType/:resourceType/:bookId',
    redirectTo: ':langId/:toolType/:resourceType/:bookId/0',
    pathMatch: 'full'
  },
  ```
  
  
I'll take a look

dr-bizz (approved):
Great work on this! This is ready to be merged in.

I think all that is left is fixes from UAT. Great work on this Caleb!

caleballdrin:
From Aaron:
Looking and working great Caleb, fantastic work on this!
Honestly I think CYOA on web is nearly done. If dismiss listener seems reasonable to get working let’s still plan to do that, but not priority and can wait if needed.

I will do the dismiss listener in another PR.

PR #249, opened 2025-04-30
Add Header to Card Collection

caleballdrin:
- Add header content
- Fix start-image styles. Currently they are only working for flex-items.

### DIFF
--- a/src/app/page/component/content-text/content-text.component.css
+++ b/src/app/page/component/content-text/content-text.component.css
@@ -12,3 +12,8 @@ p.textContent {
 :host:not(:first-child) > p.textContent {
   margin-top: 12px;
 }
+
+:host > p.hasStartImage {
+  margin: 0px;
+  width: auto;
+}
--- a/src/app/page/component/content-text/content-text.component.html
+++ b/src/app/page/component/content-text/content-text.component.html
@@ -7,7 +7,10 @@
   />
   <p
     class="mw568 textContent"
-    [ngClass]="{ firstContent: isFirstPage$ | async }"
+    [ngClass]="{
+      firstContent: isFirstPage$ | async,
+      hasStartImage: startImgResource
+    }"
     [ngStyle]="styles"
     [innerHTML]="textValue"
     dir="{{ dir$ | async }}"
--- a/src/app/page/component/content-text/content-text.component.ts
+++ b/src/app/page/component/content-text/content-text.component.ts
@@ -1,4 +1,10 @@
-import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import {
+  Component,
+  HostBinding,
+  Input,
+  OnChanges,
+  SimpleChanges
+} from '@angular/core';
 import { Observable } from 'rxjs';
 import {
   Text,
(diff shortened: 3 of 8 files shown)


canac (commented):
The code looks good, but I'm going to need help testing it unless @frett wants to. When I go to http://localhost:4200/en/tool/v2/openersstage/2/0 it just says "Failed to load the book." I'm probably doing something simple wrong.

src/app/page/component/content-text/content-text.component.ts
```diff
@@ -87,6 +95,11 @@ export class ContentTextComponent implements OnChanges {
       ? this.item.startImageSize + 'px'
       : null;
 
+    // If there is a start image, apply the alignment to the host justify content
+    this.justifyContent = this.item.startImage
```
canac, line 99: You could add a simple test for this

caleballdrin: @canac you could try running yarn again and they maybe also run this:
`yarn add @cruglobal/godtools-shared@pr790`

caleballdrin: You could also try going to the home page and then click Openers and then the Worldview category

canac: Going to http://localhost:4200/en/tool/v2/openers/2/0 works, but not http://localhost:4200/en/tool/v2/openersstage/2/0

canac, line 99: Instead of using `@HostBinding` is there a reason we can't set `styles['justify-content']` directly?

src/app/page/component/content-text/content-text.component.html
```diff
@@ -7,7 +7,10 @@
   />
   <p
```
canac, line 8: It seems that it would be possible to design this layout without the need for a `hasStartImage` class. I'm thinking you could put the image (if any) inside of the `<p>`, then apply `display: flex; gap: 0.5rem; align-items: center;` styles to the `<p>`. A couple other style tweaks might be needed, but this would make the title more flexible. You can put any elements inside of it and the alignment and layout will be correct.

Feel free to push back if I'm missing something and that wouldn't work well in this case.

canac (commented):
Let me know your thoughts about adjusting the layout to make it more flexible and not require CSS overrides when there is an image present.

src/app/page/component/content-text/content-text.component.ts
caleballdrin: We need to put justify-content on the parent element, not this component itself. @HostBinding is the only way I know how to do that.

src/app/page/component/content-text/content-text.component.html
caleballdrin: Good idea!

src/app/page/component/content-text/content-text.component.css
```diff
@@ -1,12 +1,13 @@
...
-  gap: 10px;
+  display: block;
 }
 
 p.textContent {
+  display: flex;
   margin-top: 24px;
+  flex-direction: row;
+  align-items: center;
+  gap: 0.5rem;
```
canac, line 10: The `gap` changed. I'm not sure if that matters.

src/app/page/component/content-text/content-text.component.ts
```diff
@@ -63,6 +63,7 @@ export class ContentTextComponent implements OnChanges {
         ? 'underline'
         : '',
       'text-align': this.text.textAlign.name || '',
```
canac, line 65: Do we still need the text-align style now that we have `justify-content`?

canac (approved):
Looks and tests great! 💯

Don't forget the revert the changes to `package.json` and `yarn.lock`.

PR #250, opened 2025-05-08
Remove tab labels

caleballdrin:
https://jira.cru.org/browse/GT-2607


dr-bizz (commented):
These are not appearing as tabs to me. Maybe that is the issue.

I know we have set up tabs, so I'm a little confused what why they are render as stacked content.

dr-bizz:
Ah, yes! I'm thinking of accordions.

dr-bizz (approved):

PR #253, opened 2025-05-09
Arrow and Dismiss Listener

caleballdrin:
- Add top padding on the back arrow
- Change CSS on the back arrow based on language direction
- Make close-tool listener redirect to the home page

### DIFF
--- a/src/app/page/component/page/cyoa-page/cyoa-page.component.css
+++ b/src/app/page/component/page/cyoa-page/cyoa-page.component.css
@@ -1,4 +1,17 @@
 .backwards-navigation {
-  padding: 0 20px;
+  padding-inline: 20px;
+  padding-top: 10px;
   display: inline-block;
+}
+.back-arrow-container {
+  display: flex;
+  justify-content: flex-start;
+}
+
+.back-arrow-container.rtl {
+  justify-content: flex-end;
+}
+
+.back-arrow-container.rtl img {
+  transform: rotate(180deg);
 }
--- a/src/app/page/page.component.ts
+++ b/src/app/page/page.component.ts
@@ -690,16 +690,20 @@ export class PageComponent implements OnInit, OnDestroy {
     );
 
     this.pageService.contentEvent$
-      .pipe(
-        filter((event) =>
+      .pipe(takeUntil(this._unsubscribeAll))
+      .subscribe((event) => {
+        if (event === 'close-tool') {
+          this.router.navigate(['/']);
+          return;
+        }
+
+        if (
           allListenersOnAllPages.some((allListenersOnPage) =>
             allListenersOnPage.includes(event)
           )
-        )
-      )
-      .pipe(takeUntil(this._unsubscribeAll))
-      .subscribe((event) => {
-        this.navigateToPageOnEvent(event);
+        ) {
+          this.navigateToPageOnEvent(event);
+        }
       });
   }
 
--- a/src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
+++ b/src/app/page/component/page/cyoa-card-collection-page/cyoa-card-collection-page.component.html
@@ -1,14 +1,21 @@
 <ng-container *ngIf="ready">
   <div id="toolContent">
-    <a
-      (click)="navigateBack()"
-      href="javascript:void(0)"
-      *ngIf="showBackButton"
-      class="backwards-navigation"
+    <div
+      class="back-arrow-container"
+      [ngClass]="{
+        rtl: (dir$ | async) === 'rtl',
+        ltr: (dir$ | async) === 'ltr'
+      }"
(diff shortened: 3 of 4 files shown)


src/app/page/page.component.ts
```diff
@@ -690,16 +690,20 @@ export class PageComponent implements OnInit, OnDestroy {
     );
 
     this.pageService.contentEvent$
-      .pipe(
-        filter((event) =>
+      .pipe(takeUntil(this._unsubscribeAll))
+      .subscribe((event) => {
+        if (event === 'close-tool') {
```
frett, line 695: The event you listen for should be defined on `manifest.dismissListeners`

If it's not exposed currently let me know and I'll update the exports.

frett (changes requested):

src/app/page/page.component.ts
frett: you should also be able to access it via `page.manifest.dismissListeners` depending on what objects you have available.

There should be almost nothing that is handled via a "Magic value", everything should be defined on the object model somewhere

caleballdrin: @frett I'm not able to find the 'close-tool' dismiss listener anywhere on those objects.

src/app/page/page.component.ts
```diff
@@ -690,16 +690,20 @@ export class PageComponent implements OnInit, OnDestroy {
...
             allListenersOnPage.includes(event)
           )
-        )
-      )
-      .pipe(takeUntil(this._unsubscribeAll))
-      .subscribe((event) => {
-        this.navigateToPageOnEvent(event);
+        ) {
+          this.navigateToPageOnEvent(event);
+        }
```
dr-bizz, line 706: I would be careful here, as we only want to subscribe to the events we want not all of them.
I would keep the filter, so we can filter out certain events before I subscribe to them.

src/app/page/component/page/cyoa-page/cyoa-page.component.css
```diff
@@ -1,4 +1,17 @@
 .backwards-navigation {
-  padding: 0 20px;
+  padding-inline: 20px;
+  padding-top: 10px;
```
dr-bizz, line 3:
```suggestion
  padding: 10px 20px 0;
```

dr-bizz (changes requested):
Thank you for working on this. I left some comments on how you can improve it.

src/app/page/page.component.ts
frett: It is defined on the manifest, so if you are unable to see it on the Manifest object then there is something not connected correctly

Screenshot 2025-05-12 at 10 09 54 AM

src/app/page/component/page/cyoa-page/cyoa-page.component.css
caleballdrin: I was thinking that using padding-inline and padding-top was a little more clear. Does giving only 3 values to padding make it use the middle value for both left and right?

dr-bizz: Thank you for thinking of future devs. It is fine not to use `inline`, as this is a known CSS practice and reduces the amount of CSS.

`padding: 10px;` All sides get 10px padding.
`padding: 10px 20px;` top and bottom get 10px and left and right get 20px.
`padding 10px 20px 5px` top gets 10px, left and right get 20px and bottom gets 5px.
`padding: 10px 20px 20px 5px` This would be the same as above.

src/app/page/page.component.ts
```diff
@@ -746,6 +755,17 @@ export class PageComponent implements OnInit, OnDestroy {
       }
     });
 
+    const manifestDismissListeners =
+      this._pageBookManifest?.dismissListeners || [];
+    const isManifestDismissListener = manifestDismissListeners.some(
+      (listener) => listener.name === event
+    );
+
+    if (!pageToNavigateTo && isManifestDismissListener) {
```
frett, line 764: I don't think you would need to keep the check on `pageToNavigateTo` here, closing a tool would have a higher priority than switching pages. In reality the [spec documentation](https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/manifest.xsd#L47-L49) recommends to fire off multiple events instead of relying on the same event to trigger multiple listeners

> Because multiple events can be triggered in sequence by any element that triggers events, it makes sense to prefer triggering multiple subsequent events when resolution of a single event to multiple listeners would give ambiguous results.

dr-bizz (commented):
Can you provide a way for us to test the UI? Also looks like not all the comments were addressed

dr-bizz (approved):
This works and looks amazing! Thank you for making these changes.

One thing I noticed is that it didn't redirect the user back to the homepage with their language in the URL. Can we add that before we merge?

PR #221, opened 2025-03-12
GT-2462 Add CYOA Tools to the Web App

caleballdrin:
GT-2462

- Allow CYOA type content on the web
- Re-use the tract-page component for CYOA type content
- Use text-scale to make headings in CYOA content

PR #272, opened 2025-06-13
Fix iframes and links to the old URL format

dr-bizz:
## Bug
When we deployed the code on PR https://github.com/CruGlobal/know-god-web/pull/221, we noticed that the redirect from the old URL does not work if the page number is not specified.

## Fix
We added a new redirect that redirects the old URL format of the book to the new one when the page number is not specified.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -54,6 +54,11 @@ import { SharingModalComponent } from './shared/sharing-modal/sharing-modal.comp
 
 const appRoutes: Routes = [
   {
+    path: ':langId/:bookId',
+    // Redirecting old URL format to the new one
+    redirectTo: ':langId/tool/v1/:bookId'
+  },
+  {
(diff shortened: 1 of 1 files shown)


src/app/app.module.ts
```diff
@@ -53,6 +53,11 @@ import { SharingModalComponent } from './shared/sharing-modal/sharing-modal.comp
 //============
 
 const appRoutes: Routes = [
+  {
+    path: ':langId/:bookId',
+    // Redirecting old URL format to the new one
+    redirectTo: ':langId/tool/v1/:bookId'
```
canac, line 59: I'm not familiar with the Angular router, but does this route need `pathMatch: 'full'`? Without it it seems like a `:langId/:bookId/:page` route would match both this route and the one below. I'm not sure if that will cause issues, just pointing it out.

PR #287, opened 2025-09-25
Add Copilot review instructions. Pros: improved reviews. Cons: this f…

zweatshirt:
Add Copilot review instructions. 

Pros: Improved reviews and freedom to modify as we go for different behaviors. 

Cons: this file data is sent in every chat if you use Copilot. I was hoping there was a way to avoid this but there doesn't seem to be currently. Keeping this file thin is the best avenue for now if we decide to implement it.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,13 @@
+## Reviews
+
+When reviewing pull requests:
+
+- Ensure code is correct, readable, and maintainable.
+- Identify potential bugs, security risks, or performance issues.
+- Compare new code against the style and conventions of the existing codebase.
+- Recognize and evaluate whether the PR follows established patterns.
+- When deviations occur, explain the pros and cons of the new approach.
+- Suggest specific improvements, including code examples where helpful.
+- Keep feedback clear, concise, and professional.
+- Where possible, introduce safe, relevant sources or references to support suggestions.
+- Tag your comments with the type of feedback: `bug`, `style`, `performance`, `security`, `maintainability`, `readability`, `best practice`, or `other`.


canac:
> Cons: this file data is sent in every chat if you use Copilot. I was hoping there was a way to avoid this but there doesn't seem to be currently.

To avoid filling up context, I think you can create a custom chat mode: https://code.visualstudio.com/docs/copilot/customization/custom-chat-modes#_custom-chat-modes. Then you'd select that mode and then ask it to review your code.

Also, have you tried Claude Code yet? Daniel and I have found it more effective than Copilot, even when using Claude models in Copilot.

zweatshirt:
@canac 
This file is specifically for remote GitHub reviews instead of any local behavior. I have been using Claude and I definitely prefer it to Copilot as well. 

Apparently this file works slightly different to how an `AGENT.md` or `README.md` would be handled where all the info in the file is pulled into the context. 

Here is what ChatGPT says (since the [docs](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#:~:text=in%20this%20article.-,Creating%20custom%20instructions,-You%20can%20create) were kind of ambiguous about this):

Here’s how it works under the hood:
Repository-wide instructions (your copilot-instructions.md in .github/) are always loaded whenever Copilot is running in the context of that repo.

That means Chat, autocomplete, and Reviews all can see the file.

BUT — Copilot will try to be context-aware: it looks at the headings in the file (## Chat, ## Autocomplete, ## Reviews, etc.) and uses only the relevant part for that request.

So:
If you only have ## Reviews, that section won’t be applied to Chat or autocomplete.
The file itself is still “read” at request time, but Chat won’t load irrelevant sections into its context.

PR #288, opened 2025-09-26
GT - 2760 - Separate dashboard lists and nav links

zweatshirt:
## Description

- Implements the logic to modify the view depending on what route a person is on. 
- Displays the full amount of tools or cards on when route is changed.

Related work to be done after this in a separate ticket:
- Loading indicator

### DIFF
--- a/src/app/page/component/dashboard-list/dashboard-list.component.html
+++ b/src/app/page/component/dashboard-list/dashboard-list.component.html
@@ -1,11 +1,7 @@
 <div>
   <div class="resource-header">
     <h2 class="list-header">{{ title }}</h2>
-    <a
-      href="javascript:void(0)"
-      class="view-all-link"
-      (click)="onViewResourceClick()"
-    >
+    <a *ngIf="viewAllRoute" [routerLink]="[viewAllRoute]" class="view-all-link">
       <span class="view-all-text">{{ viewAllText }} {{ title }}</span>
       <em class="far fa-arrow-right"></em>
     </a>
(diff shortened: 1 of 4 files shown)


src/app/page/component/dashboard-list/dashboard-list.component.html
```diff
@@ -2,9 +2,10 @@
   <div class="resource-header">
     <h2 class="list-header">{{ title }}</h2>
     <a
+      *ngIf="!routeSelected()"
       href="javascript:void(0)"
```
canac, line 6: A nice improvement for accessibility and Cmd+click support would be to use `routerLink` instead of a JavaScript href. That's a possible improvement for the resource cards as well.

src/app/dashboard/dashboard.component.css
```diff
@@ -18,6 +18,32 @@
...
+#homeHeader > .bs-container > .navigation {
+  display: flex;
+  gap: 20px;
+  margin-left: 20px;
+}
+
+#homeHeader > .bs-container > .navigation > a {
+  font-size: 14px;
+  color: var(--kg-gray);
+  text-decoration: none;
```
canac, line 35: Can we remove these customizations and use the default link color? (That makes the contrast a lot better too.) `text-decoration` is already `none` by default for this application's links.

```suggestion
```

zweatshirt:
@canac We briefly talked about keeping the colors the default for now so I didn't mess with the text colors. I tried to implement routerLink as you mentioned. Definitely an improvement and in my opinion less confusing code too. I did not implement it for the resource cards but not against making that change now if you'd prefer me to.

PR #290, opened 2025-10-06
GT - 2776 - Loading spinner

zweatshirt:
## Description

- Adds a simple loading spinner to the dashboard of KnowGod.com consistent with the current design

### DIFF
--- a/src/app/dashboard/dashboard.component.html
+++ b/src/app/dashboard/dashboard.component.html
@@ -28,7 +28,8 @@
 </header>
 
 <section id="homeTools">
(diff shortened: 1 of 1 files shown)


canac (commented):
There's already a loading spinner component: `<app-loader>`. What do you think about using that one?

zweatshirt:
@canac Ah yeah this is nicer. I saw the language loader and just assumed to use that. Thanks

canac (approved):
Great work! This definitely improves the UX!

PR #291, opened 2025-10-06
Lessons feature

zweatshirt:
## Description
- Creates new preview UI for Tools and Lessons
- Dedicated routes for Tools and Lessons lists
- Implements lessons functionality

TODO:
- Individual lessons impl still needs to be merged, and form logic needs to be completed

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -54,6 +54,10 @@ import { SharingModalComponent } from './shared/sharing-modal/sharing-modal.comp
 //============
 
 const appRoutes: Routes = [
+  { path: '', component: DashboardComponent },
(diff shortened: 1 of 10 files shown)


frett:
> @aaronlaib @frett This feature is ready for deployment, but I wanted to give a heads up that a few of the lessons still return a 403 code from the API side and are not accessible to users. I believe this is something that needs to be fixed on the back end but let me know if this is incorrect. The last thing we may want to update here is the godtools-shared dependency since it's still on the snapshot.

I will work on publishing an actual release of godtools-shared that we can update this PR to before we merge it. Thanks for being proactive and pointing that out @zweatshirt

PR #306, opened 2026-02-03
Stephen's dynamic content work

zweatshirt:


zweatshirt:
@smgoss Would you be able to push any changes to this branch I've created? I set the upstream of your branch on your fork to this be this repository. The deleted branch issue you mentioned as `lessons-feature` was merged into `main` a while ago.

aaronlaib:
@smgoss and @zweatshirt I just wanted to check in on this PR. Is there just the conflict on that one file (cyoa-page.component.spec.ts)?

zweatshirt:
> @smgoss and @zweatshirt I just wanted to check in on this PR. Is there just the conflict on that one file (cyoa-page.component.spec.ts)?

I haven't heard from Stephen in a bit. If I need to I'll try to take ownership of this PR soon and resolve the conflicts.

I'm out this afternoon but might be able to get to it tomorrow.

smgoss:
What is the conflict? I'm not a contributor so I can't resolve it

smgoss:
@zweatshirt 
update src/app/page/component/page/cyoa-page/cyoa-page.component.spec.ts
line 60ish to
spyOn(pageService, 'emailSignumFormDataNeeded');
spyOn(pageService, 'contentEvent');
component['init']();

That should resolve the conflict

zweatshirt:
@smgoss You should be able to push and pull to this now. I noticed some type issues still after resolving conflicts.

smgoss:
@zweatshirt When you resolved the merge conflicts, yarn regenerated the lockfile and resolved typescript@^5.4.5 to 5.9.3 (the latest matching ^5.4.5).

The upstream main branch has it pinned to 5.4.5 in its yarn.lock, but the conflict resolution blew that pin away and it needs to be downgraded to build properly. I can pin this PR to the same version but the lock file needs to be updated to resolve it.

zweatshirt:
@smgoss Hey Stephen, I'm curious how this is coming along? Let me know if there is any way I can help, maybe we can work together on it if needed if you are stuck.

smgoss:
I just got distracted and I need to fix a couple of the CI issues here and then this will be done.

smgoss:
@zweatshirt I think everything is resolved now

src/app/page/component/content-accordion/content-accordion.component.html
```diff
@@ -1,13 +1,17 @@
...
-  <ng-container *ngFor="let section of sections; trackBy: trackByFn">
-    <div class="accordion mw568">
-      <button class="accordion-header" (click)="onClick($event)">
-        {{ section.section.header.text }}
-        <i class="far fa-chevron-down"></i>
-      </button>
-      <div class="accordion-content">
-        <app-content-repeater [items]="section.contents"></app-content-repeater>
+<ng-container *ngIf="ready && !isHidden && sections">
+  <div [style.visibility]="isInvisible ? 'hidden' : 'visible'">
```
canac: question: Does this need to be two separate elements? Can we combine them? This question/suggestion applies to all of the modified components.

```suggestion
<div *ngIf="ready && !isHidden && sections" [style.visibility]="isInvisible ? 'hidden' : 'visible'">
```

src/app/page/component/content-accordion/content-accordion.component.ts
```diff
@@ -56,6 +82,22 @@ export class ContentAccordionComponent implements OnChanges {
   }
 
   private init(): void {
+    // Initialize visibility watchers
```
canac: question: This section shows up in a lot of components. Is there a way to extract it into one place instead of having to copy/paste? I'm not super familiar with Angular, but maybe inheritance could work?

src/app/page/component/content-input/content-input.component.html
```diff
@@ -1,4 +1,4 @@
-<ng-container *ngIf="ready">
+<ng-container *ngIf="ready && !isHidden">
```
canac: Does this component need an `isInvisible` check?

src/app/page/component/page/lesson-page/lesson-page.component.spec.ts
```diff
@@ -5,7 +5,7 @@ import { PageService } from '../../../service/page-service.service';
 import { LessonComponent } from './lesson-page.component';
 
 describe('LessonComponent', () => {
-  let component: LessonComponent;
+  let component: any;
```
canac: I think we can revert this change

canac (commented):
Thank you for the contribution, @smgoss! This is a solid start! I did a quick review and left some high-level feedback and possible opportunities for improvements.

kegrimes:
@canac **A couple tests:**

Family Lesson test:
- Go to `/en/lesson/lessonfamily/12`
- Ensure "Submit" button is NOT present
- Select an option & verify button appears
- Click "Submit" & check only selected option explanation appears

Neighbor Lesson test:
- Go to `/en/lesson/lessonneighbor/3`
- Ensure "Submit" button is NOT present
- Select an option & verify button appears

src/app/page/component/content-repeater/content-repeater.component.ts
```diff
@@ -56,6 +56,9 @@ export class ContentRepeaterComponent implements OnChanges {
           this.content.push({ type, content: content['content'] });
         } else if (type === 'card') {
           this.content.push({ type, content });
+        } else if (type === 'paragraph') {
+          // Keep paragraphs as paragraph components to preserve gone-if expressions
+          this.content.push({ type, content: content as ContentItems });
```
canac: Unnecessary type cast

```suggestion
          this.content.push({ type, content: content });
```

src/app/page/component/visibility-watchers/visibility-watchers.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  init(item: Content): void {
+    // Initialize visibility watchers
+    this.closeWatchers();
+
+    // Watch for gone-if expressions (removes from DOM)
+    this.isHiddenWatcher = item.watchIsGone(
+      this.state,
+      (value) => (this.isHidden = value)
```
canac: Some for the other arrow function

```suggestion
      (value) => {
        this.isHidden = value;
      }
```

canac:
@kegrimes Thanks for cleaning this up! I like how we use a service for the logic now!

Has anyone thought through the implications of changing from `<ng-container>` to a `<div>`? I think the `div` is necessary to apply inline styles, but the new element may break some of our CSS selectors and possibly layout.

From Claude:

------

**Why change made (necessary):** `ng-container` renders no element — nothing to attach `[style.visibility]` to. GodTools `invisible-if` semantics need element that hides but **keeps layout space** (`visibility: hidden`), unlike `hidden-if` which removes element entirely (`*ngIf`). So wrapper element required.

**Consequences found in this diff:**

1. **Broken CSS — concrete regression.** `content-flow-item.component.css:9`:
   ```css
   .flow-item app-content-repeater app-content-card > a { height: 100%; }
   ```
   Flow-item uses `ViewEncapsulation.None`, so rule is global and matched before. Card's `<a>` now sits inside new wrapper div → `app-content-card > a` no longer matches. Clickable cards in flow items lose `height: 100%`. Double break: even with descendant selector, new div has no height, so 100% chain snaps at the div. Same audit needed for any other direct-child (`>`) selectors crossing these component boundaries.

2. **Text component dodged this correctly.** `content-text.component.html` puts `*ngIf` + visibility binding on the `<p>` itself instead of wrapping — preserves `app-content-text > p.textContent` rules (`default-page.css:254-260`) and `:host:not(:first-child) > p.textContent` (`content-text.component.css:13`). Same with spacer (binding on existing `.spacer` div).

3. **Button/animation merged wrapper class onto new div** (`div.buttonContent`, `.animation-container`) — keeps `:host:not(:first-child) > div.buttonContent` working. Good pattern; card/link/image just nested instead.

4. **Layout box change.** Component hosts (`app-content-*`) default to `display: inline`; new div is block. Children were mostly block already (p, divs), so anonymous-box behavior largely same. Margins still collapse through bare div (no padding/border). Mostly inert — but any future `:host > *` or sibling selectors must account for extra level.

5. **Embed contract.** `visibility: hidden` content still occupies space → counted in iframe auto-resize height. Intended per invisible-if semantics, but means invisible content inflates embed height vs hidden content.

Biggest actionable item: flow-item card selector (1). Either move visibility binding onto card's existing inner elements (like text/spacer pattern) or update selector + add `height: 100%` to wrapper.

canac (commented):
Thanks for the cleanup and the verification test cases! I think after we tweak the selectors, this should be good to go!

aaronlaib:
@canac @kegrimes 

Is there any downside to me adding the on development or staging labels so I can test out on those rather than local?

canac:
> Is there any downside to me adding the on development or staging labels so I can test out on those rather than local?

@aaronlaib Go for it! It would be great to get this tested more in staging.

aaronlaib:
> @canac **A couple tests:**
> 
> Family Lesson test:
> 
> * Go to `/en/lesson/lessonfamily/12`
> * Ensure "Submit" button is NOT present
> * Select an option & verify button appears
> * Click "Submit" & check only selected option explanation appears
> 
> Neighbor Lesson test:
> 
> * Go to `/en/lesson/lessonneighbor/3`
> * Ensure "Submit" button is NOT present
> * Select an option & verify button appears

Great tests @kegrimes! The first one was one of the reasons for getting this work done as I used a good bit of the visibility logic on the Family Lesson.

Despite me being told the Staging build failed (after adding staging label), I seem to be seeing this logic working as expected on staging per your test

kegrimes:
@canac Thanks for pointing that out! Yes, `content-text` was one of the components that would break the CSS, and it was fixed by removing the `ng-container` and adding the style directly on the `<p>` element. I did the same thing with the card component but since the card template has a few different tags, I had to add the style to each one instead of a single wrapper.

I also had Claude run a sweep across the CSS for any other spots where the new element could break a selector, and the card was the only one so we should be good now!

src/app/page/component/visibility-watchers/visibility-watchers.ts
canac: Can we make this a block too?

canac (approved):
Looks great! Thanks for getting these improvements across the finish line!

src/app/page/component/visibility-watchers/visibility-watchers.ts
kegrimes: Oops, yes!

PR #313, opened 2026-05-04
GT-2868 Fix language selector issue

kegrimes:
When a user is on the tools or lessons page and desires to switch the language, it will route them to the dashboard. Instead, we want the user to remain on these pages.

Also, when on the lessons page, we only want to show the languages that have at least one lesson. 

Jira ticket: [GT-2868](https://jira.cru.org/browse/GT-2868)

### Testing
Routing Test:
- Go to `/en/lessons`
- Click on language selector and select a different language
- Ensure user remains on lessons page

Filtered Languages Test:
- Go to `/en/lessons`
- Click on language selector and select a different language
- Verify only languages with at least one lesson is present

### DIFF
--- a/src/app/dashboard/dashboard.component.spec.ts
+++ b/src/app/dashboard/dashboard.component.spec.ts
@@ -1,5 +1,11 @@
 import { HttpClientModule } from '@angular/common/http';
-import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import {
+  ComponentFixture,
+  TestBed,
+  fakeAsync,
+  tick,
+  waitForAsync
+} from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 import { mockPageComponent } from '../_tests/mocks';
 import { CommonService } from '../services/common.service';
@@ -26,6 +32,33 @@ describe('DashboardComponent', () => {
     expect(component).toBeTruthy();
   });
 
+  it('prepareLanguageSwitcher() should set availableLangs based on languagesWithLessons and current page', fakeAsync(() => {
+    component['_languagesData'] = [
+      mockPageComponent.languageEnglish,
+      mockPageComponent.languageGerman,
+      mockPageComponent.languageChineseTraditional
+    ];
+    const isLessonsPage = spyOn(component, 'isLessonsPage').and.returnValue(
+      false
+    );
+
+    // No languagesWithLessons, should include all languages
+    component.languagesWithLessons = null;
+    component['prepareLanguageSwitcher']();
+    tick();
+    expect(component.availableLangs.length).toEqual(3);
+
+    component['_languageChanged'].next();
+
+    // On lessons page with languagesWithLessons set, should filter languages
+    component.languagesWithLessons = new Set<string>(['3333']);
+    isLessonsPage.and.returnValue(true);
+    component['prepareLanguageSwitcher']();
+    tick();
+    expect(component.availableLangs.length).toEqual(1);
+    expect(component.availableLangs[0].code).toEqual('zh-Hant');
+  }));
+
   it('setDisplayLanguage() should match case-insensitively and store API correct casing', () => {
     component['_languagesData'] = [
       mockPageComponent.languageChineseTraditional
@@ -34,5 +67,29 @@ describe('DashboardComponent', () => {
     expect(component.dispLanguage).toEqual(3333);
     expect(component.dispLanguageCode).toEqual('zh-Hant');
     expect(component.dispLanguageName).toEqual('Chinese (Traditional)');
+  });
+
+  it('onSelectLanguage() should navigate to correct route based on current page', () => {
+    const navigateSpy = spyOn(component.route, 'navigate');
+    const isToolsPage = spyOn(component, 'isToolsPage').and.returnValue(false);
+    const isLessonsPage = spyOn(component, 'isLessonsPage').and.returnValue(
+      false
+    );
+
+    // On tools page
+    isToolsPage.and.returnValue(true);
+    component.onSelectLanguage('zh-Hant');
+    expect(navigateSpy).toHaveBeenCalledWith(['/', 'zh-Hant', 'tools']);
+
+    // On lessons page
+    isToolsPage.and.returnValue(false);
+    isLessonsPage.and.returnValue(true);
+    component.onSelectLanguage('zh-Hant');
+    expect(navigateSpy).toHaveBeenCalledWith(['/', 'zh-Hant', 'lessons']);
+
+    // On dashboard page
+    isLessonsPage.and.returnValue(false);
+    component.onSelectLanguage('zh-Hant');
+    expect(navigateSpy).toHaveBeenCalledWith(['/', 'zh-Hant']);
   });
 });
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -27,6 +27,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
   private readonly _lessonsRoute = 'lessons';
   tools: Resource[] = [];
   lessons: Resource[] = [];
+  languagesWithLessons: Set<string>;
   englishLangId = 1;
   englishLangCode = 'en';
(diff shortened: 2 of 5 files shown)


canac:
Thanks for working on this, Katelyn! I'm sure the GodTools will be excited to have this issue fixed!

I'm trying to delegate more things since I'm juggling a lot, so I'm going to ask @zweatshirt to review this PR.

zweatshirt:
Do you have a link to the Jira ticket by chance?

src/app/dashboard/dashboard.component.ts
```diff
@@ -27,6 +27,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
   private readonly _lessonsRoute = 'lessons';
   tools: Resource[] = [];
   lessons: Resource[] = [];
+  languagesWithLessons: Set<string>;
```
zweatshirt, line 30: I think we also want to do this for tools as well. I know the Jira ticket is kind of vague but I noticed it briefly mentions Tools

src/app/dashboard/dashboard.component.ts
```diff
@@ -82,10 +83,17 @@ export class DashboardComponent implements OnInit, OnDestroy {
...
+          .filter((lang) => {
+            if (!this.languagesWithLessons || !this.isLessonsPage()) {
+              return true;
+            }
+            return this.languagesWithLessons.has(lang.id);
+          })
+          .map((lang) => ({
+            code: lang.attributes.code,
+            name: lang.attributes.name
+          }));
```
zweatshirt, line 98: Totally optional but I might simplify this a bit to make it more readable

```suggestion
const shouldFilterByLessons =
  this.languagesWithLessons && this.isLessonsPage();

this.availableLangs = this._languagesData
  .filter(lang =>
    !shouldFilterByLessons || this.languagesWithLessons.has(lang.id)
  )
  .map(({ attributes }) => ({
    code: attributes.code,
    name: attributes.name
  }));
```

src/app/dashboard/dashboard.component.ts
```diff
@@ -182,7 +193,14 @@ export class DashboardComponent implements OnInit, OnDestroy {
...
     this.langSwitchOn = !this.langSwitchOn;
-    this.route.navigate(['/', pLangCode]);
+
+    if (this.isToolsPage()) {
+      this.route.navigate(['/', pLangCode, this._toolsRoute]);
+    } else if (this.isLessonsPage()) {
+      this.route.navigate(['/', pLangCode, this._lessonsRoute]);
+    } else {
+      this.route.navigate(['/', pLangCode]);
+    }
```
zweatshirt: I'm thinking there is a way to combine this logic with the getters below, but not entirely sure what that would look like. Not entirely necessary, I just realize we are doing some minor duplication here.

src/app/dashboard/dashboard.component.ts
```diff
@@ -182,7 +193,14 @@ export class DashboardComponent implements OnInit, OnDestroy {
     this.tools = [];
     this.lessons = [];
     this.langSwitchOn = !this.langSwitchOn;
-    this.route.navigate(['/', pLangCode]);
+
+    if (this.isToolsPage()) {
+      this.route.navigate(['/', pLangCode, this._toolsRoute]);
+    } else if (this.isLessonsPage()) {
+      this.route.navigate(['/', pLangCode, this._lessonsRoute]);
```
zweatshirt: I noticed a small issue when navigating from Tools to Lessons:

If a user has a language selected while on the Tools route, and then navigates to Lessons, there's a possibility the list will be empty since the language they selected in Tools may not have any Lessons in that language. This is minor though and I'm not entirely sure how we could best reconcile this.

zweatshirt (changes requested):
This is great Katelyn. Would we also be able to add a few tests? I know this project is unfortunately quite sparse when it comes to tests, but a couple might be good even just as self documentation of behavior for future devs. 

I'm thinking it may be good to also include filtering Tools by their available languages in this PR as well, but feel free to do that in a separate one if you'd like. I know at this point all tools should have at least one language, but it may be a good future proof.

Happy to approve once we have some tests!

src/app/dashboard/dashboard.component.ts
kegrimes: I am asking Aaron Laib in the Jira ticket!

I believe all languages have at least one tool so I don't want to do unnecessary code if it is not needed. Thanks for flagging this though!

src/app/dashboard/dashboard.component.ts
kegrimes: I think I know!

src/app/dashboard/dashboard.component.ts
kegrimes: Where are you seeing this? I don't think the list should be empty since we only filter the languages on the lessons page.

zweatshirt: @kegrimes I can show you in a huddle if you'd like.
Here are the steps:
1. The user selects 'Tools' to go to the Tools list
2. The user selects a language that has no available Lessons (e.g. Gujarati), on the Tools list.
3. The user selects 'Lessons' to go to the Lessons list.
4. The language stays the same, but because no Lessons exist for that language, it remains an empty list.

Maybe we could show simple text in that case: "No Lessons exist yet for the currently selected language.", but of course, it might be nice if that were translated. I would say don't worry about it for now, up to you.

kegrimes: Oh, are you talking about the lessons list being empty? If so, I think that should be okay because that is how GodTools behaved before. I was thinking that we could have an empty message though. Maybe something like "No lessons available"? It seems bad to just have it blank, but that would have to be an approved change. What do you think?

kegrimes: Haha, we said the same thing. Let me double check with Aaron!

src/app/services/resource.service.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+              }
+            }
+          ]
+        };
+
+        const result = service['processBookData'](mockBooksData, 1);
+
+        expect(result.tools.length).toBe(0);
+        expect(result.lessons.length).toBe(1);
+        expect(result.languagesWithLessons.size).toBe(1);
```
zweatshirt:
```suggestion
                expect(result.languagesWithLessons).toHaveSize(1);
```

src/app/services/resource.service.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      expect(service).toBeTruthy();
+    }
+  ));
+
+  describe('processBookData', () => {
+    it('should process book data correctly and exclude hidden lessons from languagesWithLessons', inject(
+      [ResourceService],
+      (service: ResourceService) => {
+        const mockBooksData = {
+          data: [
```
zweatshirt: There is a `mocks.ts` file that this may be good to move to, but since this is the only test using this mock it's probably fine for now.

src/app/dashboard/dashboard.component.spec.ts
zweatshirt, line 1: Claude suggestion (optional): One integration-style test covering the awaitBooks → languagesWithLessons → availableLangs rebuild. It's the only behavior unique to this PR that isn't exercised by the current tests.

zweatshirt (approved):
This is awesome!

src/app/services/resource.service.spec.ts
kegrimes: Good idea!

src/app/dashboard/dashboard.component.ts
kegrimes: Aaron said that it is likely that a language will have at least one tool, so I think we are okay for now. Plus, if this is something that needs added later, it should be fairly easy to replicate.

> yes, if a language appears as an option on the drop down, then at least one piece of content (either tool or lesson) has been published in that language, but most likely it will be a tool.

src/app/dashboard/dashboard.component.ts
kegrimes: There is more discussion surrounding this right now, so if anything this would be a separate PR!

PR #314, opened 2026-05-05
GT-2790 Add card style for multiselect element

kegrimes:
Add card styling on top of the already supported flat style for multiselect element. 

https://github.com/CruGlobal/mobile-content-api/blob/3bbac5c1a1b7a123e44394268713f94d2d996d94/public/xmlns/content.xsd#L989-L994 

Jira ticket: [GT-2790](https://jira.cru.org/browse/GT-2790)

**Before:**
Screenshot 2026-05-05 at 10 39 09 AM

**After:**
Screenshot 2026-05-05 at 10 39 31 AM

### Testing
- Go to `/en/lesson/lessondemogospel/6`
- Check that the new card style renders

### DIFF
--- a/src/app/page/component/content-multiselect-option/content-multiselect-option.component.css
+++ b/src/app/page/component/content-multiselect-option/content-multiselect-option.component.css
@@ -17,3 +17,8 @@
   box-shadow: inset 0px 0px 7px 2px rgba(0, 0, 0, 8%);
   opacity: 1;
 }
+
+.multiselect-option.card {
+  border-radius: 8px;
+  box-shadow: 0px 0px 7px 2px rgba(0, 0, 0, 8%);
+}
--- a/src/app/page/component/content-multiselect-option/content-multiselect-option.component.css
+++ b/src/app/page/component/content-multiselect-option/content-multiselect-option.component.css
@@ -12,13 +12,13 @@
   }
 }
 
+.multiselect-option.card {
+  border-radius: 8px;
+  box-shadow: 0px 0px 7px 2px rgba(0, 0, 0, 8%);
+}
+
 .multiselect-option.selected {
   border-radius: 8px;
   box-shadow: inset 0px 0px 7px 2px rgba(0, 0, 0, 8%);
   opacity: 1;
-}
-
-.multiselect-option.card {
-  border-radius: 8px;
-  box-shadow: 0px 0px 7px 2px rgba(0, 0, 0, 8%);
 }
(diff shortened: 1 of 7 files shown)


src/app/page/component/content-multiselect-option/content-multiselect-option.component.ts
```diff
@@ -5,13 +5,17 @@ import {
...
 import {
   Content,
   FlowWatcher,
   MultiselectOption
 } from 'src/app/services/xml-parser-service/xml-parser.service';
 import { PageService } from '../../service/page-service.service';
 
+const Style =
+  org.cru.godtools.shared.tool.parser.model.Multiselect.Option.Style;
+
```
zweatshirt: Since we are defining this in 3 separate files, we could move this to `src/app/services/xml-parser-service/xml-parser.service` and then export it from there and import it in the three files you changed: 

```
import { MultiselectOptionStyle } from 'src/app/services/xml-parser-service/xml-parser.service';
```

This is the pattern the app uses for stuff like this from the godtools module and it'll make it easier to make future changes.

src/app/page/component/content-multiselect-option/content-multiselect-option.component.css
```diff
@@ -17,3 +17,8 @@
   box-shadow: inset 0px 0px 7px 2px rgba(0, 0, 0, 8%);
   opacity: 1;
 }
+
+.multiselect-option.card {
+  border-radius: 8px;
+  box-shadow: 0px 0px 7px 2px rgba(0, 0, 0, 8%);
+}
```
zweatshirt, line 24: Claude is giving this warning as a must fix issue:
.selected and .card collide — selected card loses its selection shadow — [content-multiselect-option.component.css:15-24](vscode-webview://03s3pja2m8aahp346b0fboi3uibk62l3f5aj0u2jonm3plr802ks/src/app/page/component/content-multiselect-option/content-multiselect-option.component.css#L15-L24)


.multiselect-option.selected { border-radius: 8px; box-shadow: inset 0 0 7px 2px rgba(0,0,0,8%); }
.multiselect-option.card     { border-radius: 8px; box-shadow:        0 0 7px 2px rgba(0,0,0,8%); }
Both selectors have equal specificity (0,2,0). Because .card comes later in the file, its box-shadow wins source-order tiebreak when both classes apply. So a selected card option drops the inset shadow that signals selection — the only remaining selection feedback is the [ngStyle] selectedColor background swap.

zweatshirt (approved):
Looks good from what I can tell! My only real suggested change would be to move `const Style =
  org.cru.godtools.shared.tool.parser.model.Multiselect.Option.Style;` to `src/app/services/xml-parser-service/xml-parser.service.ts`

And the Claude suggestion may be worth looking into

src/app/page/component/content-multiselect-option/content-multiselect-option.component.ts
kegrimes: Ah, thanks for pointing that out!

PR #315, opened 2026-05-06
GT-2035 Fully support clickable interface

kegrimes:
We want to standardize support for Clickable elements so that they support firing off content events and opening a url when an item is clicked.

Components:
- Animation ✅
- Button ✅
- Card ✅ --> `en/tool/v2/openers/1`
- Image ✅
- Link 

**Jira ticket:** [GT-2035](https://jira.cru.org/browse/GT-2035)

### DIFF
--- a/src/app/page/component/content-button/content-button.component.html
+++ b/src/app/page/component/content-button/content-button.component.html
@@ -1,11 +1,30 @@
 <ng-container *ngIf="ready">
   <div class="mw568 buttonContent tc" dir="{{ dir$ | async }}">
     <a
+      *ngIf="button.url; else buttonTemplate"
       class="button"
-      [ngStyle]="{ 'background-color': buttonBgColor, color: buttonTextColor }"
+      [href]="button.url"
+      [ngStyle]="{
+        'background-color': buttonBgColor,
+        color: buttonTextColor
+      }"
+      target="_blank"
+      rel="noopener noreferrer"
       (click)="onClick()"
-      href="javascript:void(0)"
       >{{ buttonText }}</a
     >
+    <ng-template #buttonTemplate>
+      <button
+        type="button"
+        class="button"
+        [ngStyle]="{
+          'background-color': buttonBgColor,
+          color: buttonTextColor
+        }"
+        (click)="onClick()"
+      >
+        {{ buttonText }}
+      </button>
+    </ng-template>
   </div>
 </ng-container>
--- a/src/app/page/component/content-button/content-button.component.html
+++ b/src/app/page/component/content-button/content-button.component.html
@@ -1,15 +1,6 @@
 <ng-container *ngIf="ready">
   <div class="mw568 buttonContent tc" dir="{{ dir$ | async }}">
     <a
-      *ngIf="type === 'url'"
-      [href]="url"
-      [ngStyle]="{ 'background-color': buttonBgColor, color: buttonTextColor }"
-      class="button"
-      target="_blank"
-      >{{ buttonText }}</a
-    >
-    <a
-      *ngIf="type === 'event'"
       class="button"
       [ngStyle]="{ 'background-color': buttonBgColor, color: buttonTextColor }"
       (click)="formAction()"
--- a/src/app/page/component/content-button/content-button.component.html
+++ b/src/app/page/component/content-button/content-button.component.html
@@ -3,7 +3,7 @@
     <a
       class="button"
       [ngStyle]="{ 'background-color': buttonBgColor, color: buttonTextColor }"
-      (click)="formAction()"
+      (click)="onClick()"
       href="javascript:void(0)"
       >{{ buttonText }}</a
     >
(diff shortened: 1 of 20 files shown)


frett (approved):
the clickable logic looks good to me from the perspective of the spec and the design behind the different components.

I don't see anything immediately wrong with the typescript logic, I'll leave it up to you if you want @zweatshirt to take a look at the TS itself before merging

src/app/page/component/content-button/content-button.component.html
```diff
@@ -1,18 +1,9 @@
...
-      target="_blank"
-      >{{ buttonText }}</a
-    >
-    <a
-      *ngIf="type === 'event'"
       class="button"
       [ngStyle]="{ 'background-color': buttonBgColor, color: buttonTextColor }"
-      (click)="formAction()"
+      (click)="onClick()"
       href="javascript:void(0)"
```
zweatshirt: I think we will want to reinclude the anchor somehow in a follow up PR: `<a [href]>` with `target="_blank"` in cases of urls.

`href="javascript:void(0)"` is an accessibility anti-pattern and we also lose benefits like the ability to open links in a new tab.

With that, I am thinking we should also do this with the other clickable elements.

I asked Claude what that might look like:

> ```
> <a
>   *ngIf="link.url; else buttonTpl"
>   [href]="link.url"
>   target="_blank"
>   rel="noopener noreferrer"
>   (click)="onClick()"
> >{{ linkText }}</a>
> <ng-template #buttonTpl>
>   <button type="button" (click)="onClick()">{{ linkText }}</button>
> </ng-template>
> ```
> Then in handleClickable, drop the window.open(url, '_blank') line — the anchor handles navigation natively. onClick only needs to fire events.

Since this will likely cause duplication, we can try putting this in a template if you think it'll work well.

src/app/page/component/content-card/content-card.component.spec.ts
```diff
@@ -49,7 +47,23 @@ describe('ContentCardComponent', () => {
     const pageService = TestBed.get(PageService);
     spyOn(pageService, 'formAction');
 
-    component.eventClick();
+    component.onClick();
```
zweatshirt: I think we may want to modify this test, since it is similar to the test just below but not immediately obvious. Here, `window.open` is being called as well, we just aren't spying on and asserting it. 

I'd either delete this test or preferably test the "events only" and "url only" paths.

src/app/page/component/content-image/content-image.component.spec.ts
```diff
@@ -48,29 +50,72 @@ describe('ContentImageComponent', () => {
   });
 
   it('correctly calls events', () => {
```
zweatshirt: Similar situation in this test. I don't believe this is just triggering events, but the url handler is also called. Wondering if we could improve the tests a bit?

src/app/page/component/content-link/content-link.component.spec.ts
```diff
@@ -30,7 +30,6 @@ describe('ContentInputComponent', () => {
 
     expect(component.text).toEqual(link.text);
     expect(component.linkText).toBe('link text');
-    expect(component.events).toEqual(link.events);
   });
 
   it('Form Action function sends the correct info to pageService', async () => {
```
zweatshirt: Same situation as mentioned in the other comments I made.

zweatshirt (approved):
I think this works well from what I can tell, thanks Katelyn! Because of that I am happy to approve, but also hoping we can address some of the review comments now or at a later date. None are blocking but still important.

kegrimes:
@zweatshirt Can you take a quick glance at my changes since you last reviewed? I want to make sure that what I did is what you were thinking!

I looked through the UI and found animations, buttons, and images and they appear to be working the same as before. However, I could not find an example where cards or links were used.

src/app/page/component/content-card/content-card.component.html
```diff
@@ -7,11 +7,24 @@
...
   </a>
+  <ng-template #cardTemplate>
+    <a
+      *ngIf="card.isClickable"
+      class="card clickable"
+      (click)="onClick()"
+      [ngStyle]="{ 'background-color': background }"
+    >
+      <app-content-repeater [items]="contents"></app-content-repeater>
+    </a>
```
zweatshirt, line 28: question: Do we want to eventually make this a button as well? I'm hesitant because then we would have to fix the styling. But it would technically give users ability to tab on cards. I would probably avoid adding to this PR at this point though

zweatshirt:
@kegrimes Here is an example of cards being used: https://knowgod.com/en/tool/v2/openers/1

zweatshirt (approved):
Great work Katelyn. I tried to run through the UI to notice any regressions and couldn't catch any. Everything is looking good.

src/app/page/component/content-card/content-card.component.html
kegrimes: Yeah, this could probably be saved for the future, but good call out!

PR #317, opened 2026-05-13
GT-2031 Refactor image prefetcher

kegrimes:
Currently, the image pre-fetcher is grabbing all resources instead of the ones that are needed. We need to refactor this function to filter out images and animations not included in the manifest.

Jira ticket: [GT-2031](https://jira.cru.org/browse/GT-2031)

**Technical Solution:**
Each manifest has a `relatedFiles` field that is an array of strings. These strings are `sha256` of resources plus a file extension, which we will use to compare to every resource:
```
"2eeb2bca42e959fe74ab23ae9f330474a323c239fa2aee6c82476daf8f3d82f4.xml"
"78c7cdab8e6cfc2782df1dd25fb447bf0f8fb4eea76aa4593a13d6c5cdab6ccf.jpg"
"85f00dabd27410cd08d23c8f6c733f80edea388af4ba875835c88ce6422253e2.png"
```

`.xml` are pages, so these won't be included; `.jpg` and `.png` are images and will be included if they are needed in the current manifest

Resource data:
```
{
    "id": "1069",
    "type": "attachment",
    "attributes": {
        "file": "https://mobile-content-api.cru.org/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsiZGF0YSI6NzkzLCJwdXIiOiJibG9iX2lkIn19--806555080cc5ea19387e07c8471848da0304bb65/four_En_01.jpg",
        "is-zipped": false,
        "sha256": "78c7cdab8e6cfc2782df1dd25fb447bf0f8fb4eea76aa4593a13d6c5cdab6ccf",
        "file-file-name": "four_En_01.jpg"
    },
    "relationships": {
        "resource": {
            "data": {
                "id": "6",
                "type": "resource"
            }
        }
    }
},
```

**Testing:**
You can `console.log`...
- The filenames of attachment types in `pageBookIndex` to see all possible resources
- Manifest's `relatedFiles` to see all the files the manifest uses
- The filtered resources after both checks to see the resources that are actually loaded (after line 390)

### DIFF
--- a/src/app/page/page.component.spec.ts
+++ b/src/app/page/page.component.spec.ts
@@ -4,12 +4,17 @@ import { Router } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
 import {
   createEventId,
+  mockPageBookIndexData,
   mockPageComponent,
   mockTractPage
 } from '../_tests/mocks';
 import { CommonService } from '../services/common.service';
 import { LoaderService } from '../services/loader-service/loader.service';
-import { godToolsParser } from '../services/xml-parser-service/xml-parser.service';
+import {
+  ManifestParser,
+  XmlParserData,
+  godToolsParser
+} from '../services/xml-parser-service/xml-parser.service';
 import { PageComponent, getResourceTypeEnum } from './page.component';
 import { PageService } from './service/page-service.service';
 
@@ -277,6 +282,49 @@ describe('PageComponent', () => {
         );
       });
     });
+  });
+
+  it('loadBookManifestXML() fetches only needed resources', async () => {
+    component._pageBookIndex = mockPageBookIndexData;
+
+    component._selectedLanguage = { id: '2222' };
+    component._pageBookTranslations = [
+      {
+        id: 'tx-1',
+        attributes: { 'manifest-name': 'foo.xml' },
+        relationships: { language: { data: { id: '2222' } } }
+      }
+    ];
+
+    const fakeManifest = {
+      relatedFiles: {
+        asJsReadonlySetView: () => new Set(['aaa111.png', 'bbb222.png'])
+      },
+      pages: []
+    };
+    spyOn(ManifestParser.prototype, 'parseManifest').and.returnValue(
+      Promise.resolve({ manifest: fakeManifest } as unknown as XmlParserData)
+    );
+
+    const addAttachmentSpy = spyOn(pageService, 'addAttachment');
+    spyOn(component, 'getResource');
+
+    component['loadBookManifestXML']();
+    await fixture.whenStable();
+
+    expect(addAttachmentSpy).toHaveBeenCalledTimes(2);
+    expect(addAttachmentSpy).toHaveBeenCalledWith(
+      'needed-1.png',
+      'https://cru.org/needed-1.png'
+    );
+    expect(addAttachmentSpy).toHaveBeenCalledWith(
+      'needed-2.png',
+      'https://cru.org/needed-2.png'
+    );
+    expect(addAttachmentSpy).not.toHaveBeenCalledWith(
+      'skipped.png',
+      'https://cru.org/skipped.png'
+    );
   });
 
   it('getAvailableLanguagesForSelectedBook() no languages downloaded', () => {
--- a/src/app/page/page.component.spec.ts
+++ b/src/app/page/page.component.spec.ts
@@ -307,23 +307,24 @@ describe('PageComponent', () => {
     );
 
     const addAttachmentSpy = spyOn(pageService, 'addAttachment');
-    spyOn(component, 'getResource');
+    const getResourceSpy = spyOn(component, 'getResource');
 
     component['loadBookManifestXML']();
     await fixture.whenStable();
 
-    expect(addAttachmentSpy).toHaveBeenCalledTimes(2);
-    expect(addAttachmentSpy).toHaveBeenCalledWith(
-      'needed-1.png',
-      'https://cru.org/needed-1.png'
+    expect(addAttachmentSpy).toHaveBeenCalledTimes(3);
+
+    expect(getResourceSpy).toHaveBeenCalledWith(
+      getResourceTypeEnum.image,
+      'needed-1.png'
     );
-    expect(addAttachmentSpy).toHaveBeenCalledWith(
-      'needed-2.png',
-      'https://cru.org/needed-2.png'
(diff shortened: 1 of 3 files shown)


src/app/page/page.component.ts
```diff
@@ -367,25 +371,38 @@ export class PageComponent implements OnInit, OnDestroy {
...
-                  : getResourceTypeEnum.animation,
-                attributes['file-file-name']
-              );
+
+            if (type !== 'attachment') {
+              return;
+            }
+            if (!neededShas.has(attributes.sha256)) {
+              return;
             }
```
zweatshirt: Very opinionated but I might combine these if statements. It will still short circuit as long as the SHAs check comes after the attachment check

src/app/page/page.component.spec.ts
```diff
@@ -279,6 +284,49 @@ describe('PageComponent', () => {
...
+      Promise.resolve({ manifest: fakeManifest } as unknown as XmlParserData)
+    );
+
+    const addAttachmentSpy = spyOn(pageService, 'addAttachment');
+    spyOn(component, 'getResource');
+
+    component['loadBookManifestXML']();
+    await fixture.whenStable();
+
+    expect(addAttachmentSpy).toHaveBeenCalledTimes(2);
```
zweatshirt: What do you think of removing this assertion? the assertions beneath it seem sufficient here

src/app/page/page.component.spec.ts
```diff
@@ -279,6 +284,49 @@ describe('PageComponent', () => {
...
+    ];
+
+    const fakeManifest = {
+      relatedFiles: {
+        asJsReadonlySetView: () => new Set(['aaa111.png', 'bbb222.png'])
+      },
+      pages: []
+    };
+    spyOn(ManifestParser.prototype, 'parseManifest').and.returnValue(
+      Promise.resolve({ manifest: fakeManifest } as unknown as XmlParserData)
```
zweatshirt, line 306: If it's not too much of a refactor I am thinking we can avoid this unknown typecast

zweatshirt (changes requested):
This is looking great so far. I noticed that some images actually fail to load. For example http://localhost:4200/en/tool/v1/poweroverfear/1 
- Compare against https://knowgod.com/en/tool/v1/poweroverfear/1 

Here is the error:
```
poweroverfear locale: en resource: fp_web_1-min-min.png la: Legacy Manifest missing Image Resource tool: poweroverfear locale: en resource: fp_web_1-min-min.png
```

I asked Claude about this,

> The new filter at [page.component.ts:388](vscode-webview://03s3pja2m8aahp346b0fboi3uibk62l3f5aj0u2jonm3plr802ks/src/app/page/page.component.ts#L388) skips both addAttachment AND getResource (which calls addToImagesDict) for any attachment whose SHA isn't in relatedFiles. Since fp_web_1-min-min.png's SHA isn't in relatedFiles (because it's not in \<resources>\), the dict never gets populated for it.

Let me know if you'd like me to take a further look tomorrow and I will.

src/app/page/page.component.ts
kegrimes: I actually need to keep these separate now since I needed to fix some of the images breaking!

src/app/page/page.component.spec.ts
kegrimes: I may just keep it since there is another test file that uses the unknown typecast (`cyoa-card-collection-page.component.spec.ts`). It may not be ideal, but not sure if it is 100% worth it. Feel free to disagree though!

kegrimes:
@zweatshirt Those console errors were existing before, but thank you for finding that issue with the images! 

Turns out `addAttachment` was the function populating the lookup table that pages use to find their images at render time and some pages reference images that aren't listed in the manifest. So, I needed to call it for every attachment to keep the lookup table complete, and only use the actual prefetch with my filtered resources!

frett:
My few comments:
- I actually just changed what `relatedFiles` returns, it now returns an array of objects since each file could have a sha256 checksum and size associated with it.
   - When the godtools-shared library is upgraded in this project we will address this functional change then.
- There are also lottie animations present in the `relatedFiles`, those files are currently `.json`. I'm not sure if that applies to the image pre-loader or not, but wanted to make you aware.
- Regarding the legacy image warning/error, there is a [legacy image behavior](https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/blob/main/module/parser/src/commonMain/kotlin/org/cru/godtools/shared/tool/parser/model/Image.kt#L46-L50) used by web that generates `Resource` objects for images that historically were used web only and not included in the Manifest. These placeholder image objects would not be present in `relatedFiles` since the manifest doesn't know about them. If we can identify which tools have those images then a republish of that tool should fix the warning and the mising image

kegrimes:
@frett 
1. Thank you for letting me know! I can leave a TODO comment in the code so we can come back to it when the godtools-shared library is upgraded. Should I create a Jira ticket?
2. Yes, I already took into account the animations! We just have a general check that will remove the extensions so we can compare it to the `sha256` resources field in the manifest.
3. Here is the list of tools that have the error present: (I looked through all the English tools)

Power over Fear
- fp_web_1-min-min.png
- fp_web_2-min-min.png
- fp_web_3-min-min.png
- fp_web_4-min-min.png
- fp_web_5-min-min.png
- fp_web_6-min-min.png

Knowing God Personally
- Web-kgp-point-1.png
- Web-kgp-point-2.png
- Web-kgp-everyone-is-sinful.png
- Web-kgp-graphic-sin-has-consequences-graphic-person-small.png
- Web-kgp-point-3.png
- Web-kgp-jesus-died-in-our-place.png
- Web-kgp-hands.png
- web-kgp-graphic-god-proves-his-love-person-small.png
- Web-kgp-point-4.png
- Web-kgp-Self-Directed.png
- Web-kgp-Christ-Directed.png
- Web-kgp-prayer.png
- Web-kgp-unsure.png
- Web-kgp-plane.png
- Web-kgp-growth-6.png

zweatshirt (approved):
This works well, great work Katelyn!

I am a little concerned with future drift if the parser changes down the road, but this is a good change.

frett:
Drift down the road is expected, but should be minimal. The library is versioned and I try to avoid breaking changes as much as possible. So know God can always stay on its current library version until we are ready to upgrade

I also have some validation workflows setup to identify when a change will impact iOS or JS

PR #318, opened 2026-05-15
GT-3001 Setup crowdin

kegrimes:
We want to setup Crowdin so we can translate the hardcoded strings in the frontend! Most logic was pulled from how MPDX react has Crowdin setup.

The secret has been added via Github UI

Jira ticket: [GT-3001](https://jira.cru.org/browse/GT-3001)

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -84,8 +84,7 @@ jobs:
         with:
           node-version-file: .tool-versions
       - name: 📦 Install Dependencies
-        # Download packages from the registry instead of git-lfs to save git-lfs bandwidth
-        run: yarn cache clean && yarn install
+        run: yarn install
 
       - name: 🌐 Extract Translations
         run: yarn extract
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -75,6 +75,32 @@ jobs:
       - name: 🏗️ Build app
         run: yarn build
 
+  translations:
+    name: 🌐 Translations
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .tool-versions
+      - name: 📦 Install Dependencies
+        # Download packages from the registry instead of git-lfs to save git-lfs bandwidth
+        run: yarn cache clean && yarn install
+
+      - name: 🌐 Extract Translations
+        run: yarn extract
+      - name: 🔤🔼 CrowdIn Translations Upload
+        uses: crowdin/github-action@v2
+        # Always run `yarn extract` to verify labels but only push translations to CrowdIn when PRs merge
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        with:
+          upload_sources: true
+          upload_translations: false
+          download_sources: false
+          download_translations: false
+        env:
+          CROWDIN_API_TOKEN: ${{ secrets.CROWDIN_API_TOKEN }}
+
   deploy:
     name: 🚀 Deploy app
     if: (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/development')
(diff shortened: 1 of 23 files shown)


.github/workflows/node.js.yml
```diff
@@ -75,6 +75,32 @@ jobs:
...
+    name: 🌐 Translations
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .tool-versions
+      - name: 📦 Install Dependencies
+        # Download packages from the registry instead of git-lfs to save git-lfs bandwidth
+        run: yarn cache clean && yarn install
```
dr-bizz:
```suggestion
      - name: 📦 Install dependencies
        run: yarn install
```

dr-bizz: do we need yarn cache clean?

src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -7,6 +7,7 @@ import {
   waitForAsync
 } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
+import { I18NextModule } from 'angular-i18next';
```
dr-bizz, line 10: CAn we check that we need to import `I18NextModule`. I don't think we do as we've imported it globally on app.module.ts

src/app/services/xml-parser-service/xml-parser.service.ts
```diff
@@ -226,8 +226,9 @@ export const parseTextRemoveBrTags = (text: string): string => {
 @Injectable({
   providedIn: 'root'
 })
-export class PullParserFactory extends org.cru.godtools.shared.tool.parser.xml
-  .JsXmlPullParserFactory {
+export class PullParserFactory
```
dr-bizz, line 229: Did the linter just update this? Seems like nothing changed.

embed/test.html
```diff
@@ -12,8 +12,8 @@
       }
 
       body {
-        font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
-          sans-serif;
+        font-family:
```
dr-bizz, line 15: Did the linter just update this? Seems like nothing changed.

crowdin.yml
```diff
@@ -0,0 +1,14 @@
+'project_id': '897600'
+'api_token_env': CROWDIN_API_TOKEN
+'base_path': '.'
+'base_url': 'https://api.crowdin.com'
+'pull_request_labels': ['crowdin']
+'pull_request_reviewers': ['dr-bizz', 'canac']
```
dr-bizz: Lets add you, Zach and Will to this list

i18next-parser.config.cjs
```diff
@@ -0,0 +1,39 @@
+const EventEmitter = require('events');
```
dr-bizz, line 1: Did you end up using this?

dr-bizz (changes requested):

dr-bizz (changes requested):
We have a few things to iron out, but it's mostly all there. Have you manually added the translations to CrowdIn? so the phrases are there already?

embed/test.html
kegrimes: Yes, this was the linter!

src/app/services/xml-parser-service/xml-parser.service.ts
kegrimes: Yes, this was the linter!

.github/workflows/node.js.yml
kegrimes: No we don't. I missed that from copying from MPDX project

src/app/dashboard/dashboard.component.spec.ts
kegrimes: If we don't include this import, we get this error: `Error: NG0302: The pipe 'i18next' could not be found in the 'DashboardComponent' component.`

i18next-parser.config.cjs
kegrimes: Yes, in line 3. If we don't use it, we get this error: `TypeError: lexer.on is not a function`

kegrimes:
@dr-bizz I have not and I'm not sure how to do it. Would you mind showing me?

dr-bizz (approved):

PR #328, opened 2026-06-10
GT-2855 Adjust flow item behvaior

kegrimes:
## Description

Currently for flow items, `gone-if` is being removed correctly, but still retaining position. To fix this, I have moved the visibility logic up a level to the parent (`content-flow-item` to `content-flow`) so it can own each watcher per item. 

Jira ticket: [GT-2855](https://jira.cru.org/browse/GT-2855)

**Before:**
Screenshot 2026-06-10 at 4 54 48 PM

**After:**
Screenshot 2026-06-10 at 4 54 13 PM

## Testing

- Go to `/en/tool/v1/emojitool/1`
- Select 3 random emojis and go to next page
- Check that items and position are removed

## Checklist:

- [x] I have requested a review from another person on the project
- [x] I have tested my changes on staging or development

### DIFF
--- a/src/app/page/component/content-flow/content-flow.component.html
+++ b/src/app/page/component/content-flow/content-flow.component.html
@@ -7,10 +7,12 @@
   >
     <!-- Subtract 17.5px from the percentage to account for the gap between flow-items  -->
     <app-content-flow-item
-      *ngFor="let item of items; trackBy: trackByFn"
-      [item]="item"
+      *ngFor="let contentItem of items; trackBy: trackByFn"
+      [item]="contentItem.item"
+      [hidden]="contentItem.visibility?.isHidden"
       [ngStyle]="{
-        flexBasis: 'calc(' + item.itemWidth + ' - 17.5px)'
+        flexBasis: 'calc(' + contentItem.itemWidth + ' - 17.5px)',
+        visibility: contentItem.visibility?.isInvisible ? 'hidden' : 'visible'
       }"
     >
     </app-content-flow-item>
--- a/src/app/page/component/content-flow/content-flow.component.ts
+++ b/src/app/page/component/content-flow/content-flow.component.ts
@@ -1,29 +1,42 @@
-import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import {
+  Component,
+  Input,
+  OnChanges,
+  OnDestroy,
+  SimpleChanges
+} from '@angular/core';
 import {
   DimensionParser,
   Flow,
   FlowItem,
-  Horizontal,
-  ParserState
+  Horizontal
 } from 'src/app/services/xml-parser-service/xml-parser.service';
 import { PageService } from '../../service/page-service.service';
+import { VisibilityWatchers } from '../visibility-watchers/visibility-watchers';
+
+type FlowContent = {
+  item: FlowItem;
+  itemWidth?: string | null;
+  visibility?: VisibilityWatchers;
+};
 
 @Component({
   selector: 'app-content-flow',
   templateUrl: './content-flow.component.html',
   styleUrls: ['./content-flow.component.css']
 })
-export class ContentFlowComponent implements OnChanges {
+export class ContentFlowComponent implements OnChanges, OnDestroy {
   @Input() item: Flow;
 
   flow: Flow;
-  items: (FlowItem & { itemWidth?: string | null })[];
+  items: FlowContent[];
   ready: boolean;
-  state: ParserState;
   justifyContent: Horizontal;
 
-  constructor(private pageService: PageService) {
-    this.state = this.pageService.parserState();
+  constructor(private pageService: PageService) {}
+
+  ngOnDestroy(): void {
+    this.closeItemWatchers();
   }
 
   ngOnChanges(changes: SimpleChanges) {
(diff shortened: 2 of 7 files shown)


src/app/page/component/content-flow/content-flow.component.ts
```diff
@@ -49,13 +61,22 @@ export class ContentFlowComponent implements OnChanges {
...
     this.items = this.item.items;
     this.items.forEach((flowItem) => {
       const dimension = DimensionParser(flowItem.width);
       flowItem.itemWidth = dimension?.value
         ? dimension.value + dimension.symbol
         : null;
+
+      const visibility = new VisibilityWatchers(this.pageService);
+      visibility.init(flowItem);
+      flowItem.visibility = visibility;
```
kegrimes: [Medium] **Watcher lifecycle state stored on the shared `FlowItem` model.** `flowItem.visibility = visibility` writes a subscription-bearing `VisibilityWatchers` onto a godtools-shared parsed model object, and overwrites any existing one **without closing it first**. `closeItemWatchers()` only reaches items still in `this.items`, so if the parser ever hands back a `FlowItem` instance that already carries a `.visibility` (a shared/reused node), the previous watcher's two `FlowWatcher`s leak. Low real-world likelihood (flows re-parse fresh) → Medium, but the fix is cheap and matches the repo convention (every other component owns its watcher in a private field, never on the model).

```ts
private itemWatchers: VisibilityWatchers[] = [];

private closeItemWatchers(): void {
  this.itemWatchers.forEach((w) => w.closeWatchers());
  this.itemWatchers = [];
}
// in init() forEach:
const visibility = new VisibilityWatchers(this.pageService);
visibility.init(flowItem);
flowItem.visibility = visibility;     // keep for the template read
this.itemWatchers.push(visibility);   // own the cleanup handle
```

Minimal alternative: add `flowItem.visibility?.closeWatchers();` immediately before the overwrite. _Flagged by Reactive (6), Architecture (5), Testing (2)._

src/app/page/component/content-flow/content-flow.component.html
```diff
@@ -9,8 +9,10 @@
     <app-content-flow-item
       *ngFor="let item of items; trackBy: trackByFn"
       [item]="item"
+      [hidden]="item.visibility?.isHidden"
```
kegrimes: [Medium] **`[hidden]` for the gone case deviates from the documented contract.** The 15 sibling content templates use `*ngIf` (removes from DOM) for gone, and `visibility-watchers.ts` literally comments *"Watch for gone-if expressions (removes from DOM)"* — but `[hidden]` only sets `display:none`, leaving the element in the DOM. The agents agree this is **deliberate and correct here** (keeping the host rendered lets the parent-owned watcher keep observing so a gone item can later un-hide, and keeps flex layout stable). Two cheap follow-ups: (a) add a one-line comment on this binding explaining *why* `[hidden]` and not `*ngIf`, so a future "simplification" doesn't re-break it; (b) the "removes from DOM" comment in `visibility-watchers.ts` is now slightly misleading for this usage. _Flagged by Standards (5), Reactive (3)._

src/app/page/component/content-flow/content-flow.component.ts
```diff
@@ -49,13 +61,22 @@ export class ContentFlowComponent implements OnChanges {
...
   private init(): void {
+    this.closeItemWatchers();
     this.items = this.item.items;
     this.items.forEach((flowItem) => {
       const dimension = DimensionParser(flowItem.width);
       flowItem.itemWidth = dimension?.value
         ? dimension.value + dimension.symbol
         : null;
+
+      const visibility = new VisibilityWatchers(this.pageService);
```
kegrimes, line 80: [Suggestion] **New per-item watcher orchestration is untested.** `content-flow.component.spec.ts` exists but only asserts `items`/`ready` — it doesn't cover that a watcher is created per item, that gone/invisible state flows through, or that `closeItemWatchers()` runs on destroy and re-init. The codebase convention is to test visibility behavior (`visibility-watchers.spec.ts`, `content-text-filtering.spec.ts`). A 4-test skeleton (create-per-item, reflect gone/invisible, close-on-destroy, close-on-reinit) using the `wireWatchers` spy pattern is available on request. Note: `mockFlow()` items default `watchIsGone: () => null`, so tests must spy before `ngOnChanges`. _Flagged by Testing (4)._

kegrimes (commented):
## 🤖 Multi-Agent Code Review — PR #328

**Verdict: ✅ APPROVED WITH SUGGESTIONS** — no blockers.

Five specialized agents (Architecture, Reactive & Async, UX/Accessibility, Testing, Standards) + dependency analysis reviewed this PR. All five independently agreed the central design — moving visibility from `content-flow-item` up to the host-owning parent `content-flow` — is **correct**: the flex item is the `<app-content-flow-item>` host, which the child can't collapse, so the leaf-component visibility pattern legitimately doesn't apply here. The child simplification and the per-item watcher lifecycle (`closeItemWatchers()` on both `ngOnDestroy` and re-init) are sound.

**Dependency impact: LOW** — `app-content-flow-item` is rendered only by `content-flow`; no external reads of the removed `visibility` field; no breaking changes.

### Findings (all non-blocking)
| # | Severity | File | Theme |
|---|----------|------|-------|
| 1 | 6.0 Medium | content-flow.component.ts:79 | Watcher lifecycle stored on shared model / overwrite-without-close |
| 2 | 5.0 Medium | content-flow.component.html:12 | `[hidden]` for gone deviates from documented "removes from DOM" contract |
| 3 | 4.0 Suggestion | content-flow.component.ts:77 | New per-item watcher orchestration is untested |

**Recommended before merge:** Finding #1 — own the watchers in a component-private array (or guard the overwrite with `flowItem.visibility?.closeWatchers()`). A few lines; makes the lifecycle bulletproof rather than "fine in practice." #2 is a clarifying comment; #3 has a ready-to-paste 4-test skeleton (ask if you want it).

**UX agent confirmed** `[hidden]` actually collapses the flex slot — no `display` rule on `app-content-flow-item` overrides the UA `[hidden]{display:none}`, and a `display:none` flex child reserves no `gap`. The fix works at the CSS level; gone vs invisible semantics are correct.

> ⚠️ Risk score is **9/10 CRITICAL**, but it's driven entirely by the file-pattern weighting (4 × `src/app/page/component/**` files × +2). The actual change is a small, self-contained layout fix — weigh the CRITICAL label with that context.

_Note: this review does not approve the PR (event: COMMENT). Reply `/dismiss: <reason>` on any comment below to dismiss it (severity < 7 only)._

canac (approved):
I think this looks good, and it tests great too! Thanks for the reproduction URL! That made it really easy to test.

PR #333, opened 2026-07-20
[GT-2823] - Add search input to navbar language selector

wjames111:
## Description

Related Jira ticket: [GT-2823](https://jira.cru.org/browse/GT-2823)

- Adds a search input to both navbar language selectors: the dashboard language modal and the tool-page header dropdown.
- Adds a shared helper (`src/app/shared/language-search.ts`) that builds a search key per language from the API display name, the language code, and `Intl.DisplayNames`-derived names: the language's own name for itself (e.g. "Deutsch", "español") and its name localized to the currently displayed language (e.g. "ingles" finds English while browsing in Spanish). This makes search work across languages regardless of how the list is displayed, per the ticket's note about displayed language text varying.
- Matching is case- and diacritic-insensitive ("francais" matches "Français"); language codes `Intl` has no data for fall back gracefully to name + code matching.
- Shows a "No languages found" empty state; the query is cleared each time a selector is opened. Search inputs use `dir="auto"` so RTL queries render correctly.
- Adds two i18n keys (`Search languages`, `No languages found`) to `src/assets/locales/en/translation.json`.
- Unit tests added for the helper and for filtering/reset behavior in both components (197 specs passing).

No dependent PRs.

## Testing

- Run `yarn start:dev` and go to `http://localhost:4200/en`
- Click the language name in the top-right of the navbar to open the language modal
- Type "espanol" — check that only "Spanish" remains; type "francais" — check the French variants remain; type gibberish — check the "No languages found" message appears
- Clear the search and open a tool (e.g. go to `/en/tool/v1/90minutes/0`), then click the language selector in the tool header
- Repeat the searches above in the dropdown (this tool has fewer languages, so e.g. "deutsch" correctly shows no results); try "العربية" — check "Arabic" matches
- Click a filtered language (e.g. Spanish) — check the tool reloads in that language; then search "ingles" — check "English" matches while the UI language is Spanish
- Close and reopen either selector — check the previous query is cleared

## Checklist:

- [ ] I have requested a review from another person on the project
- [ ] I have tested my changes on staging or development

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app/dashboard/dashboard.component.html
+++ b/src/app/dashboard/dashboard.component.html
@@ -78,14 +78,26 @@
     <div class="modal-close">
       <i class="fas fa-times" (click)="onSwitchLanguage()"></i>
     </div>
+    <div class="language-search">
+      <input
+        type="search"
+        dir="auto"
+        [(ngModel)]="languageSearchText"
+        [placeholder]="'Search languages' | i18nextEager"
+        [attr.aria-label]="'Search languages' | i18nextEager"
+      />
+    </div>
     <ul class="languages-grid">
       <li
         class="language"
-        *ngFor="let lang of availableLangs"
+        *ngFor="let lang of filteredLangs"
         (click)="onSelectLanguage(lang.code)"
       >
         {{ lang.name }}
       </li>
     </ul>
+    <div class="no-language-results" *ngIf="filteredLangs.length === 0">
+      {{ 'No languages found' | i18nextEager }}
+    </div>
   </div>
 </div>
--- a/src/app/dashboard/dashboard.component.html
+++ b/src/app/dashboard/dashboard.component.html
@@ -96,8 +96,15 @@
         {{ lang.name }}
       </li>
     </ul>
-    <div class="no-language-results" *ngIf="filteredLangs.length === 0">
-      {{ 'No languages found' | i18nextEager }}
+    <div
+      class="no-language-results"
+      role="status"
+      aria-live="polite"
+      [class.has-message]="filteredLangs.length === 0"
+    >
+      <ng-container *ngIf="filteredLangs.length === 0">
+        {{ 'No languages found' | i18nextEager }}
+      </ng-container>
     </div>
   </div>
 </div>
--- a/src/app/dashboard/dashboard.component.html
+++ b/src/app/dashboard/dashboard.component.html
@@ -80,9 +80,11 @@
     </div>
     <div class="language-search">
       <input
+        #languageSearchInput
         type="search"
         dir="auto"
-        [(ngModel)]="languageSearchText"
+        [ngModel]="languageSearchText"
+        (ngModelChange)="onLanguageSearchTextChange($event)"
         [placeholder]="'Search languages' | i18nextEager"
         [attr.aria-label]="'Search languages' | i18nextEager"
       />
@@ -100,9 +102,9 @@
       class="no-language-results"
       role="status"
       aria-live="polite"
-      [class.has-message]="filteredLangs.length === 0"
+      [class.has-message]="showNoLanguagesMessage"
     >
-      <ng-container *ngIf="filteredLangs.length === 0">
+      <ng-container *ngIf="showNoLanguagesMessage">
         {{ 'No languages found' | i18nextEager }}
       </ng-container>
     </div>
--- a/src/app/dashboard/dashboard.component.spec.ts
+++ b/src/app/dashboard/dashboard.component.spec.ts
@@ -9,6 +9,7 @@ import {
 import { FormsModule } from '@angular/forms';
 import { RouterTestingModule } from '@angular/router/testing';
 import { I18NextModule } from 'angular-i18next';
+import { NEVER } from 'rxjs';
 import { mockPageComponent } from '../_tests/mocks';
 import { CommonService } from '../services/common.service';
 import { DashboardComponent } from './dashboard.component';
@@ -66,6 +67,19 @@ describe('DashboardComponent', () => {
     expect(component.availableLangs[0].code).toEqual('zh-Hant');
   }));
 
+  it('prepareLanguageSwitcher() should ignore languagesReady emissions before languages load', fakeAsync(() => {
+    spyOn(component.commonService, 'getLanguages').and.returnValue(NEVER);
+    component['prepareLanguageSwitcher']();
+
+    // awaitBooks re-emits languagesReady when dashboard data arrives, which
+    // can happen before the languages request resolves.
+    component['_languagesReady'].next();
+    tick();
+
+    expect(component.availableLangs).toBeUndefined();
+    expect(component.filteredLangs).toEqual([]);
+  }));
+
   it('filteredLangs should filter by search text', fakeAsync(() => {
     component['_languagesData'] = [
       mockPageComponent.languageEnglish,
@@ -77,22 +91,33 @@ describe('DashboardComponent', () => {
     component['prepareLanguageSwitcher']();
     tick();
 
-    component.languageSearchText = '';
     expect(component.filteredLangs.length).toEqual(3);
+    expect(component.showNoLanguagesMessage).toBeFalse();
 
     // Matches the name shown in the selector
-    component.languageSearchText = 'germ';
+    component.onLanguageSearchTextChange('germ');
     expect(component.filteredLangs.map((l) => l.code)).toEqual(['de']);
 
-    component.languageSearchText = 'no such language';
+    component.onLanguageSearchTextChange('no such language');
     expect(component.filteredLangs.length).toEqual(0);
+    expect(component.showNoLanguagesMessage).toBeTrue();
   }));
 
-  it('onSwitchLanguage() should clear the search text', () => {
-    component.languageSearchText = 'germ';
+  it('onSwitchLanguage() should clear the search text and restore the full list', fakeAsync(() => {
+    component['_languagesData'] = [mockPageComponent.languageGerman];
+    component.languagesWithLessons = null;
+    spyOn(component, 'isLessonsPage').and.returnValue(false);
+    component['prepareLanguageSwitcher']();
+    tick();
+
+    component.onLanguageSearchTextChange('no such language');
+    expect(component.filteredLangs.length).toEqual(0);
+
     component.onSwitchLanguage();
     expect(component.languageSearchText).toEqual('');
-  });
+    expect(component.filteredLangs.length).toEqual(1);
+    tick();
+  }));
 
   it('setDisplayLanguage() should match case-insensitively and store API correct casing', () => {
     component['_languagesData'] = [
--- a/src/app/dashboard/dashboard.component.spec.ts
+++ b/src/app/dashboard/dashboard.component.spec.ts
@@ -6,6 +6,7 @@ import {
   tick,
   waitForAsync
 } from '@angular/core/testing';
+import { FormsModule } from '@angular/forms';
 import { RouterTestingModule } from '@angular/router/testing';
 import { I18NextModule } from 'angular-i18next';
 import { mockPageComponent } from '../_tests/mocks';
@@ -19,7 +20,12 @@ describe('DashboardComponent', () => {
   beforeEach(waitForAsync(() => {
     TestBed.configureTestingModule({
       declarations: [DashboardComponent],
-      imports: [HttpClientModule, RouterTestingModule, I18NextModule.forRoot()],
+      imports: [
+        HttpClientModule,
+        FormsModule,
+        RouterTestingModule,
+        I18NextModule.forRoot()
+      ],
       providers: [CommonService]
     }).compileComponents();
   }));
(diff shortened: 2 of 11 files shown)


src/app/page/page.component.spec.ts
```diff
@@ -126,6 +132,37 @@ describe('PageComponent', () => {
...
+    expect(component.filteredLanguages.length).toEqual(2);
+
+    // Matches the name shown in the selector
+    component.languageSearchText = 'eng';
+    expect(component.filteredLanguages).toEqual([
+      mockPageComponent.languageEnglish
+    ]);
+
+    // Matches the language's own name for itself
+    component.languageSearchText = 'Deutsch';
```
wjames111: [Suggestion] This assertion depends on the browser's ICU/CLDR data (`Intl.DisplayNames` returning exactly `'Deutsch'`). The utility spec (`language-search.spec.ts`) is the right home for ICU assertions — there they're load-bearing. Here the intent is "search filters the list," which the `'eng'` case already proves with app-controlled data. Consider replacing `'Deutsch'` with an API-name query so a stripped-ICU Chrome or CLDR retranslation can't fail this spec for an unrelated reason.

src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -60,6 +66,38 @@ describe('DashboardComponent', () => {
...
+
+    component.languageSearchText = '';
+    expect(component.filteredLangs.length).toEqual(3);
+
+    // Matches the name shown in the selector
+    component.languageSearchText = 'germ';
+    expect(component.filteredLangs.map((l) => l.code)).toEqual(['de']);
+
+    // Matches the language's own name for itself
+    component.languageSearchText = 'Deutsch';
```
wjames111: [Suggestion] Same ICU-data dependency as the page spec: asserting `'Deutsch'` couples this component spec to the browser's CLDR data. The `'germ'` case above already proves filtering with app-controlled input; the endonym path is covered in `language-search.spec.ts` where it belongs.

src/app/page/page.component.html
```diff
@@ -36,16 +36,31 @@
...
               dir="ltr"
               (click)="selectLanguage(lang, true)"
-              *ngFor="let lang of availableLanguages"
+              *ngFor="let lang of filteredLanguages"
             >
               {{ lang.attributes.name }}
             </li>
           </ul>
+          <div
+            class="no-language-results"
```
wjames111: [Suggestion] Screen readers get no feedback as results filter or this empty state appears — consider `aria-live="polite"` on this div (WCAG 4.1.3 Status Messages, AA). Note: the debate concluded this is subordinate to the pre-existing click-only `<li>` items (keyboard users can't select results at all), so the two are best fixed together in a follow-up rather than blocking this PR. Same applies to the dashboard modal.

src/app/dashboard/dashboard.component.html
```diff
@@ -78,14 +78,26 @@
...
       <li
         class="language"
-        *ngFor="let lang of availableLangs"
+        *ngFor="let lang of filteredLangs"
         (click)="onSelectLanguage(lang.code)"
       >
         {{ lang.name }}
       </li>
     </ul>
+    <div class="no-language-results" *ngIf="filteredLangs.length === 0">
```
wjames111: [Suggestion] Same as the page-header dropdown: an `aria-live="polite"` region here would announce the filtered/empty state to screen-reader users. Best paired with making the `<li>` results keyboard-operable (pre-existing gap) in a follow-up ticket.

src/app/page/page.component.ts
```diff
@@ -233,6 +239,33 @@ export class PageComponent implements OnInit, OnDestroy {
...
+    }
+    return this.availableLanguages.filter((lang) =>
+      languageMatchesSearch(
+        this.getLanguageSearchKey(lang),
+        this.languageSearchText
+      )
+    );
+  }
+
+  private getLanguageSearchKey(lang: Language): string {
```
wjames111, line 280: [Suggestion] Three small follow-ups on this cache (none blocking — the debate confirmed the locale-keyed Map is the *correct* design here, since `availableLanguages` holds shared model objects this component shouldn't mutate):
1. Clear `_languageSearchKeys` in `clearData()` — it's the one Map not reset there (`_visibleHiddenPageIds.clear()` sets the convention). Correctness already survives via the `|${langId}` key; this just bounds growth.
2. Add a spec for the cache-hit branch (call `filteredLanguages` twice, spy that `buildLanguageSearchKey` runs once) and one asserting a locale change mints a distinct entry.
3. A one-line comment noting the shared-model constraint would stop future refactors from "unifying" this with the dashboard's eager approach, which is eager only because it owns its `{code,name,searchKey}` DTOs.

wjames111 (commented):
## Multi-Agent Code Review — CLEAN ✅

**Verdict: CLEAN** — 0 blockers, 0 important issues. 5 specialized agents (Security, Architecture, Testing, UX, Standards) + dependency analysis, with cross-examination debate and rebuttal rounds.

**Risk: LOW (1/10)** — self-contained UI feature; no dependency, auth, schema, or workflow changes.

### Highlights
- **Security:** clean — safe Angular bindings throughout, no RegExp built from user input (plain `String.includes`), static `dir` attributes, embed iframe auto-resize contract untouched.
- **Standards:** all checklist items PASS (shared-helper placement, plain co-located CSS, Prettier/import order, testing conventions, no godtools-shared overlap).
- **Dependency impact:** no breaking changes; `language-search.ts` has exactly its 3 intended importers; the `availableLangs` shape change is additive with no external readers.
- Every initial concern (rated 6.0–6.5) was downgraded or retracted during debate once challenged with code evidence — e.g. the "empty state flashes during load" concern was retracted after tracing both load sequences, and the per-change-detection filtering concern was conceded to be ~123 `String.includes` calls per tick with all `Intl.DisplayNames` work memoized outside the getter.

### Minor suggestions not posted as line comments (severity ≤ 3.5, optional)
- Add a direct assertion that the filter getters return `[]` when the backing list is undefined.
- The native `type="search"` clear button (`::-webkit-search-cancel-button`) is unstyled and will vary across browsers.
- The empty state could offer a one-tap "clear search" affordance.
- A one-line comment explaining the `getIntlLanguageName(code, code)` self-endonym trick would help future readers.

---

### Pre-existing issue (informational — predates this PR, does not affect the verdict)

**[Pre-existing, severity 8.0] Language list items are click-only** (`page.component.html` / `dashboard.component.html`): the `<li>` results have no `tabindex`, `role`, or keydown handling, so keyboard and screen-reader users can focus the new search input but cannot select a result. This pattern dates to 2023 and is not regressed by this PR — but the search input is now the first focusable control in the flow, which surfaces it. Recommend a follow-up ticket pairing keyboard operability (`role="option"`, `tabindex`, Enter/Space + arrow keys) with the `aria-live` suggestion below.

src/app/page/page.component.ts
```diff
@@ -233,6 +239,33 @@ export class PageComponent implements OnInit, OnDestroy {
 
   onToggleLanguageSelect(): void {
     this.languagesVisible = !this.languagesVisible;
+    this.languageSearchText = '';
+  }
+
+  get filteredLanguages(): Language[] {
```
wjames111, line 253: [Suggestion] This getter runs on every change-detection tick while the dropdown is open (twice per tick counting the empty-state guard), re-normalizing the query once per language. At 123 languages this is microseconds and consistent with existing template getters, so it's fine as-is — but hoisting the query normalization out of the loop (or recomputing the filtered list on `ngModelChange`) would be a cheap in-idiom tidy-up. Same note applies to `filteredLangs` in the dashboard.

src/app/page/page.component.spec.ts
wjames111: 🤖 Fixed in 98ac338e: Replaced the ICU-dependent 'Deutsch' query with the app-controlled 'german' (matches the mock's attributes.name)

_Posted by Claude Code via /agent-fix._

src/app/dashboard/dashboard.component.spec.ts
wjames111: 🤖 Fixed in 48f0cdf8: Removed the ICU-dependent 'Deutsch' assertion; the app-controlled 'germ' case already covers filtering, endonym matching stays covered in language-search.spec.ts

_Posted by Claude Code via /agent-fix._

src/app/page/page.component.html
wjames111: 🤖 Fixed in 2f53a300: Made .no-language-results a persistent role=status aria-live=polite region with the message toggled on an inner span (padding moved to the span so the empty region leaves no gap)

_Posted by Claude Code via /agent-fix._

src/app/dashboard/dashboard.component.html
wjames111: 🤖 Fixed in 223d16b9: Same persistent aria-live=polite region for the modal; padding moved to a .has-message modifier so the empty div leaves no visible gap

_Posted by Claude Code via /agent-fix._

src/app/page/page.component.ts
wjames111: 🤖 Fixed in bfa87c8d: Cleared _languageSearchKeys in clearData(), added cache-hit and per-locale specs (via Intl.DisplayNames spy), and documented the shared-model constraint above getLanguageSearchKey

_Posted by Claude Code via /agent-fix._

src/app/page/page.component.ts
wjames111: 🤖 Fixed in f6085dc7: Added a shared filterLanguages helper that normalizes the query once per filter pass (early-returns on empty query); both getters now use it, with specs

_Posted by Claude Code via /agent-fix._

canac:
I see some layout shift when filtering. Can we fix that?

https://github.com/user-attachments/assets/dc14193c-5b9f-4587-98b8-d7dbcd86cbc5

src/app/dashboard/dashboard.component.ts
canac, line 129: I'm seeing a TypeError in DevTools from this line:

```
dashboard.component.ts:199 ERROR TypeError: Cannot read properties of undefined (reading 'filter')
    at Object.next (dashboard.component.ts:110:12)
```

src/app/shared/language-search.ts
```diff
@@ -0,0 +1,66 @@
...
+    return displayNames.of(code) || null;
+  } catch {
+    return null;
+  }
+}
+
+export function buildLanguageSearchKey(
+  code: string,
+  name: string,
+  displayLocale?: string
```
canac: Does this need to be optional? It is always passed a value.

src/app/shared/language-search.ts
```diff
@@ -0,0 +1,66 @@
...
+): string {
+  const names = [name, code, getIntlLanguageName(code, code)];
+  if (displayLocale) {
+    names.push(getIntlLanguageName(code, displayLocale));
+  }
+  return names.filter(Boolean).map(normalizeLanguageText).join('\n');
+}
+
+export function languageMatchesSearch(
```
canac: Is this method used? It looks like dead code.

src/app/page/page.component.html
```diff
@@ -36,16 +36,30 @@
       </div>
       <div id="languageList" class="py25 ng-hide" *ngIf="languagesVisible">
         <div class="mw568">
+          <div class="language-search">
+            <input
```
canac, line 40: Should we autofocus the search input when it becomes visible?

src/app/dashboard/dashboard.component.html
```diff
@@ -78,14 +78,33 @@
...
     </ul>
+    <div
+      class="no-language-results"
+      role="status"
+      aria-live="polite"
+      [class.has-message]="filteredLangs.length === 0"
+    >
+      <ng-container *ngIf="filteredLangs.length === 0">
+        {{ 'No languages found' | i18nextEager }}
+      </ng-container>
```
canac, line 109: This "No languages found" state renders whenever `filteredLangs` is empty for any reason — including while `availableLangs` is still loading (the getter returns `[]` until data arrives). Because this container is `aria-live="polite"`, screen readers announce a false empty state during that initial load even though languages are about to appear. Consider gating the message on a non-empty search term (or on the list having finished loading) so it only shows for a genuine no-match.

src/app/dashboard/dashboard.component.ts
```diff
@@ -109,7 +114,12 @@ export class DashboardComponent implements OnInit, OnDestroy {
           })
           .map(({ attributes }) => ({
             code: attributes.code,
-            name: attributes.name
+            name: attributes.name,
+            searchKey: buildLanguageSearchKey(
```
canac: `buildLanguageSearchKey` constructs two `Intl.DisplayNames` instances per language, and this runs for the whole list on every `_languagesReady` emission — which fires at least twice on a normal load (initial prepare, then again from `awaitBooks` after the dashboard data arrives). That is a lot of upfront main-thread work for the full language list even when the user never opens the switcher. The page component builds these keys lazily; could we do the same here (or defer until the modal opens)?

src/app/dashboard/dashboard.component.ts
```diff
@@ -203,6 +213,18 @@ export class DashboardComponent implements OnInit, OnDestroy {
 
   onSwitchLanguage(): void {
     this.langSwitchOn = !this.langSwitchOn;
+    this.languageSearchText = '';
+  }
+
+  get filteredLangs(): { code: string; name: string; searchKey: string }[] {
```
canac: `filteredLangs` is evaluated three times per change-detection cycle (the `*ngFor` at line 93 plus the two `.length === 0` checks at lines 103 and 105), and getters are not memoized, so `filterLanguages` re-runs over the full list each time. Storing the result in a field that updates when `languageSearchText` changes would compute it once per change instead.

src/app/page/page.component.ts
```diff
@@ -233,6 +239,35 @@ export class PageComponent implements OnInit, OnDestroy {
...
+    );
+  }
+
+  // Keys live in an external Map because availableLanguages holds shared
+  // parsed Language models this component must not mutate (unlike the
+  // dashboard, which owns its mapped objects and can build keys eagerly).
+  private getLanguageSearchKey(lang: Language): string {
+    const cacheKey = `${lang.id}|${this._pageParams.langId}`;
+    let searchKey = this._languageSearchKeys.get(cacheKey);
+    if (!searchKey) {
```
canac: The cache guard `if (!searchKey)` treats an empty-string key as a miss, so any language whose normalized key is `""` gets recomputed (two `Intl.DisplayNames` constructions) on every read. `if (!this._languageSearchKeys.has(cacheKey))` would be the correct check. Unlikely with real data, but the guard is technically wrong.

src/app/dashboard/dashboard.component.html
```diff
@@ -78,14 +78,33 @@
     <div class="modal-close">
       <i class="fas fa-times" (click)="onSwitchLanguage()"></i>
     </div>
+    <div class="language-search">
+      <input
```
canac, line 82: Should we autofocus the input when it becomes visible?

canac (approved):
This is amazing! Much easier to navigate now. It's cool that you can search in English or in the native language. Approving, pending the comments I posted about. Let's get Aaron's feedback too before merging to production.

src/app/dashboard/dashboard.component.ts
wjames111: Didn't see this, I'll try to get rid of it.

src/app/page/page.component.html
wjames111: Yeah I think that would be a good idea.

PR #334, opened 2026-07-22
[GT-3052] - Add start/end-image rendering in button element

wjames111:
## Description

- Renders the button's `icon` attribute before/after the label per `icon-gravity` (`icon-size` as pixel width), matching the mobile renderer ([RenderButton.kt](https://github.com/CruGlobal/kotlin-mpp-godtools-tool-parser/blob/main/module/renderer/src/commonMain/kotlin/org/cru/godtools/shared/renderer/content/RenderButton.kt#L70-L78)). Previously buttons rendered only their plain text label, so icons showed on mobile but not on web (e.g. the buttons at the end of a lesson).
- Icons only render for `icon-gravity` `start`/`end` (CENTER never renders, mirroring mobile), are vertically centered with 8px RTL-safe `margin-inline` spacing, and resolve URLs with the same images-dict + attachment-fallback logic the text component uses.
- Accessibility: icons stay decorative (`alt=""`) when the button has a text label; an icon-only button gets a non-empty alt fallback (icon name, else `button`) so the control keeps an accessible name (WCAG 4.1.2).
- Per review (@frett): `start-image`/`end-image` are defined on standalone text elements only in the content schema, so buttons intentionally do **not** render them — a spec pins that they're ignored on the button's child text node.
- The `<a>` and `<button>` variants share one inner `ng-template` so both render identically.

Jira: [GT-3052](https://jira.cru.org/browse/GT-3052)

Verified live against staging content: the end-of-lesson buttons (e.g. `lessonfamily` page 16 "Additional resources" with `icon-gravity="end"`) render their icons at the authored `icon-size`.

## Testing

- Run `yarn start:dev` and open http://localhost:4200/en/lessonfamily/16 (or `lessonneighbor` page 13)
- Check the "Additional resources" / "Replay lesson" / "Close lesson" buttons render their icon after the label, sized per `icon-size` and vertically centered (note: the replay/close icons are blue-on-blue until button `style="outlined"` support lands — a separate ticket)
- Check buttons without icons render exactly as before (no stray ``, unchanged spacing)
- In an RTL language, check the icon margins flip with the text direction
- `yarn test --no-watch` — 190/190 pass, including specs for icon start/end/CENTER gravity, the anchor variant, attachment fallback, alt behavior, and that text-node start/end images are ignored

## Checklist:

- [ ] My PR is ready for review and I have applied the **"Review"** label — this automatically requests a review from a randomly-chosen team member
- [ ] I have tested my changes on staging or development

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/app/page/component/content-button/content-button.component.css
```diff
@@ -9,4 +9,17 @@ div.buttonContent {
 
 :host:not(:first-child) > div.buttonContent {
     margin-top: 12px;
+}
+
+.button img.buttonImg {
+    vertical-align: middle;
+    max-height: 46px;
```
wjames111: [Medium] `max-height: 46px` fights the inline `width` the template forces on these images. Because width is fixed while height is clamped, an image whose proportional height exceeds 46px (e.g. `start-image-size` 100+ on a squarish image) renders distorted. The sibling components that render the same model images (`content-text`, `content-image`) size by width only with no height cap — matching mobile, which lets height follow the width.

Realistic content sizes (icons ~18px, text images ~40px) don't hit the cap, so this is not a blocker — but consider either dropping the cap to match the siblings, or swapping it for narrow-embed hardening:

```css
.button img.buttonImg {
    vertical-align: middle;
    max-width: 100%;
}
```

src/app/page/component/content-button/content-button.component.html
```diff
@@ -27,7 +28,37 @@
...
+      <ng-container *ngTemplateOutlet="buttonContent"></ng-container>
     </button>
   </ng-template>
+  <ng-template #buttonContent>
+    <img
+      *ngIf="iconResource && iconGravity === 'START'"
+      class="buttonImg buttonImgStart"
+      [src]="iconResource"
+      [ngStyle]="{ width: iconWidth }"
+      alt=""
```
wjames111: [Medium] Accessible-name gap for image-only buttons (WCAG 4.1.2). `alt=""` is the right call while `buttonText` is present (matches the `content-image` decorative convention) — but if content authors a button whose `Text` has images with an empty string, or an icon-only button (the icon renders regardless of `button.text`), the `<a>`/`<button>` ends up with no accessible name.

Since both branches share this `#buttonContent` template, one edit covers both — e.g. bind a non-empty `alt` on the icon only when the label is empty, or add `[attr.aria-label]` on the hosts when `buttonText` is empty and an image/icon is present.

src/app/page/component/content-button/content-button.component.html
```diff
@@ -27,7 +28,37 @@
       }"
       (click)="onClick()"
     >
-      {{ buttonText }}
+      <ng-container *ngTemplateOutlet="buttonContent"></ng-container>
     </button>
   </ng-template>
+  <ng-template #buttonContent>
+    <img
+      *ngIf="iconResource && iconGravity === 'START'"
```
wjames111, line 38: [Medium] The `iconGravity === 'START' | 'END'` guards mean a `CENTER` gravity icon (the parser default) intentionally renders nothing — mirroring mobile — but no spec pins that behavior, and it's the branch most likely to silently regress. One cheap assertion covers it:

```ts
it('does not render the icon when gravity is CENTER', () => {
  const pageService = TestBed.inject(PageService);
  pageService.addToImagesDict('icon.png', 'https://cru.org/icon.png');
  const b = {
    ...mockButton(buttonText, '', buttonEvent),
    icon: createResource('icon.png', 'icon.png'),
    iconGravity: { name: 'CENTER', ordinal: 1 },
    iconSize: 18
  } as Button;
  component.item = b;
  component.ngOnChanges({ item: new SimpleChange(null, b, true) });
  fixture.detectChanges();
  expect(fixture.nativeElement.querySelectorAll('button img').length).toBe(0);
});
```

src/app/page/component/content-button/content-button.component.spec.ts
```diff
@@ -100,4 +101,116 @@ describe('ContentButtonComponent', () => {
...
+      text: mockText(buttonText)
+    } as Button;
+    component.item = buttonWithTextImages;
+    component.ngOnChanges({
+      item: new SimpleChange(null, buttonWithTextImages, true)
+    });
+    fixture.detectChanges();
+
+    const images: NodeListOf<HTMLImageElement> =
+      fixture.nativeElement.querySelectorAll('button img');
```
wjames111: [Suggestion] All five new image tests query `button img`, so the `<a href>` variant with images is never asserted. Divergence risk is near-zero since both branches render the same shared `#buttonContent` template — but one test building a URL button (`mockButton(buttonText, buttonUrl, '')` + `text: mockText(...)`) and asserting `querySelectorAll('a img').length` would lock the outlet wiring on the anchor side.

src/app/page/component/content-button/content-button.component.ts
```diff
@@ -75,8 +90,32 @@ export class ContentButtonComponent implements OnChanges, OnDestroy {
...
     this.ready = true;
   }
+
+  private resolveImage(resource: Resource | null): string | null {
+    if (!resource?.name) {
+      return null;
+    }
+    let url = this.pageService.getImageUrl(resource.name);
+    // Try to find image in all attachments
+    if (url === resource.name && !url.includes('http')) {
```
wjames111: [Suggestion] Two follow-up notes on `resolveImage` (no change needed in this PR):

1. `!url.includes('http')` skips the attachment fallback for any image *named* with an "http" substring (e.g. `httpsign.png`), letting the bare name reach `src`. `startsWith('http')` is the precise form — but this is inherited verbatim from `content-text`, so changing it here alone would diverge.
2. This is now the third copy of the getImageUrl→findAttachment fallback (`content-text.component.ts:90-100` and `content-image` inline it) — and the cleanest one. A follow-up could promote this helper to `PageService` and migrate all three, fixing the guard in one place.

wjames111 (commented):
## Multi-Agent Code Review (standard mode)

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers; 3 medium-priority improvements recommended.

**Risk:** LOW (1/10) · 4 review agents (Architecture, Testing, Standards, UX) + dependency analyzer, with cross-examination debate and rebuttal rounds.

### Summary
- **0 critical / 0 high / 0 important** findings. 3 medium (5.0–6.9) posted as line comments, plus 2 suggestions (informational).
- **Dependency impact:** no breaking changes. The `mocks.ts` change (plain-object `text` on the button mock) was verified against all 33 specs importing `_tests/mocks` — none rely on Kotlin `Text` behavior; the swap prevents `No manifest found in model ancestors` crashes now that the component reads `startImage`/`endImage`.
- **Verified clean:** embed contract (height re-measured via MutationObserver + delayed re-sends in `src/index.html`; auto-resize is height-only), RTL (`margin-inline-*` + inline flow), empty-string `src` can never render (`*ngIf` + `|| null` guard), mock swap safe across consumers, all CLAUDE.md standards PASS (imports, prettier, co-located CSS/specs, yarn-only, no debug output), mobile-parity gravity semantics (icon only for START/END) correct.
- **Debate note:** an initial challenge claimed `width` + `max-height` preserves aspect ratio per CSS 2.1 §10.4 — overruled: the ratio-preserving constraint table applies only when both dimensions are auto; the template forces an inline `width`, so the clamp does distort oversized images (realistic content sizes 18–40px don't hit the 46px cap, hence medium severity).

### Test coverage suggestions (informational, severity < 5)
- **[4.0]** `resolveImage` total-miss path (dict miss + attachment miss → no ``) — verified correct, just unasserted.
- **[4.0]** falsy `startImageSize`/`iconSize` → no width style — unasserted.
- **[3.5]** full ordering with icon + both text images together — unasserted.

Suggestions (severity < 5) are informational and do not require `/dismiss`. Findings with severity < 7 can be dismissed by replying `/dismiss: <reason>` on the comment.

src/app/page/component/content-button/content-button.component.css
wjames111: 🤖 Fixed in a985043e: replaced `max-height: 46px` with `max-width: 100%` so button images size by width only, matching content-text/content-image.

_Posted by Claude Code via /agent-fix._

src/app/page/component/content-button/content-button.component.html
wjames111: 🤖 Fixed in 61520efd: added an `imgAlt` fallback — images stay decorative (`alt=""`) when the button has a text label, otherwise use the icon/start/end resource name so the control keeps an accessible name; spec added for the icon-only case.

_Posted by Claude Code via /agent-fix._

src/app/page/component/content-button/content-button.component.html
wjames111: 🤖 Fixed in 75f108e3: added a spec pinning that a CENTER-gravity icon (parser default) renders no img.

_Posted by Claude Code via /agent-fix._

src/app/page/component/content-button/content-button.component.spec.ts
wjames111: 🤖 Fixed in ff68f266: added a spec asserting both text images render inside the anchor (`a href`) variant.

_Posted by Claude Code via /agent-fix._

wjames111 (commented):
## Multi-Agent Code Review — Re-run after review fixes (standard mode)

**Verdict: APPROVED WITH SUGGESTIONS** — 0 blockers. All findings from the previous review are verified fixed.

**Risk:** LOW (1/10) · 4 fresh review agents (Architecture, Testing, Standards, UX), full PR diff at head `ff68f26`.

### Prior findings — verified resolved
- **Image sizing** — `max-width: 100%` swap confirmed: images scale by width with preserved aspect ratio, matching `content-text`/`content-image` and mobile; embed height contract re-verified intact (MutationObserver + delayed height re-sends in `src/index.html`).
- **Accessible name** — `imgAlt` fallback satisfies WCAG 4.1.2 for text-less buttons; confirmed Angular's `[alt]="''"` renders a present-but-empty `alt=""` (correct decorative semantics). The `'icon'` extension-strip assertion is correct.
- **Coverage gaps** — the CENTER-gravity and anchor-variant specs were independently re-derived as correct by fresh agents.

### Quality gates (run during review)
`yarn test --no-watch` → **189/189 pass** · `yarn prettier:check` → clean · `yarn lint` → 0 errors (9 pre-existing warnings in unrelated files).

### Remaining suggestions
Two line comments below (max severity 5.0, none block merge): the decorative `alt=""` branch lacks an assertion, and `imageAltFallback()` has minor polish opportunities (CENTER-gravity gate, untested fallback branches, unlocalized filename alt as follow-up ticket material).

Suggestions with severity < 7 can be dismissed by replying `/dismiss: <reason>`; severity < 5 items are informational.

src/app/page/component/content-button/content-button.component.spec.ts
```diff
@@ -100,4 +104,182 @@ describe('ContentButtonComponent', () => {
...
+    expect(images.length).toBe(2);
```
wjames111: [Medium] The *decorative* side of the new alt branch (`imgAlt = ''` when `buttonText` is present) renders in these image tests but is never asserted — only the fallback side has coverage. Two one-liners here pin it:

```ts
expect(images[0].getAttribute('alt')).toBe('');
expect(images[1].getAttribute('alt')).toBe('');
```

src/app/page/component/content-button/content-button.component.ts
```diff
@@ -75,8 +92,46 @@ export class ContentButtonComponent implements OnChanges, OnDestroy {
...
+    // them a non-empty alt when they are the only content so the control
+    // still has an accessible name (WCAG 4.1.2).
+    this.imgAlt = this.buttonText ? '' : this.imageAltFallback();
+
     this.ready = true;
   }
+
+  private imageAltFallback(): string {
+    const name =
+      this.button.icon?.name ||
```
wjames111: [Suggestion] Three polish notes on `imageAltFallback()` (none blocking — the `'button'` default keeps WCAG 4.1.2 satisfied in every branch):

1. It prefers `button.icon?.name` even when the icon won't render (CENTER gravity), so a text-less button's visible start/end image can get an invisible icon's filename as its alt. One-line gate:
```ts
const gravity = this.button.iconGravity?.name;
const iconName =
  gravity === 'START' || gravity === 'END' ? this.button.icon?.name : null;
```
2. The `startImage`/`endImage` name path and the ultimate `'button'` default are untested (only the icon path is asserted).
3. Filename-derived alt (e.g. `circle_blue`) is unlocalized — fine as a fallback; sourcing a content-authored localized label would be the real fix and pairs well with the existing PageService-consolidation follow-up.

src/app/page/component/content-button/content-button.component.spec.ts
wjames111: 🤖 Fixed in b8e34c8d: added the two alt="" assertions to the start/end-image test, pinning the decorative branch when the button has a text label.

_Posted by Claude Code via /agent-fix._

src/app/page/component/content-button/content-button.component.ts
wjames111: 🤖 Fixed in 73b3cdd4: imageAltFallback() now only uses the icon name when the icon actually renders (START/END gravity), with new specs covering the text-image fallback and the ultimate "button" default. Point 3 (localized content-authored alt) stays with the PageService-consolidation follow-up.

_Posted by Claude Code via /agent-fix._

src/app/page/component/content-button/content-button.component.html
```diff
@@ -27,7 +28,37 @@
...
+      class="buttonImg buttonImgStart"
+      [src]="iconResource"
+      [ngStyle]="{ width: iconWidth }"
+      [alt]="imgAlt"
+    />
+    <img
+      *ngIf="startImgResource"
```
canac: question: The Kotlin snippet from the ticket only renders icons with `iconGravity`, right? Is support for `startImgResource` and `endImgResource` needed for the web? Maybe mobile just doesn't support those two yet.

src/app/page/component/content-button/content-button.component.ts
```diff
@@ -75,8 +92,52 @@ export class ContentButtonComponent implements OnChanges, OnDestroy {
...
+  private imageAltFallback(): string {
+    // Only prefer the icon name when the icon actually renders (START/END
+    // gravity); CENTER icons are never displayed.
+    const iconName =
+      this.iconGravity === 'START' || this.iconGravity === 'END'
+        ? this.button.icon?.name
+        : null;
+    const name =
+      iconName ||
+      this.button.text?.startImage?.name ||
```
canac: If the button has a start image and an end image, I think this could fallback to using the start image alt text for both images if `button.icon` doesn't have a name.

canac (approved):
Looks really good from a code standpoint and the one page that I looked at! We can get Aaron to QA and see if it's all working how he expected.

src/app/page/component/content-button/content-button.component.html
frett: the intention is to not include start-image & end-image from the text node, I believe I have that restricted in the xml schema. I'm going to double check this.

Aaron referencing `start-image`/`end-image` was probably just him referencing the wrong name for this ticket

frett: `start-image`/`end-image` is defined on [standalone text elements only](https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/content.xsd#L586-L611)

the button type has it's own [child text element](https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/content.xsd#L786-L792) which just implements the [`textType`](https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/content.xsd#L627-L667) type, which does not include those attributes.

frett (changes requested):
buttons don't need to support start/end images on the child text nodes, they only need to support icon & icon gravity defined on the button itself.

src/app/page/component/content-button/content-button.component.html
wjames111: 🤖 Confirmed and fixed in b2b6bf9a — @frett is right: per the content schema, `start-image`/`end-image` exist on standalone text elements only, so I removed the text-image rendering entirely. Buttons now render images solely via `icon`/`icon-gravity`/`icon-size`, matching the mobile renderer. A spec now pins that start/end images on the child text node are ignored, and the button mock workaround in `_tests/mocks.ts` is reverted (no longer needed). 190/190 tests pass.

_Posted by Claude Code._

src/app/page/component/content-button/content-button.component.ts
wjames111: 🤖 Moot as of b2b6bf9a — start/end text images are no longer rendered in buttons (schema allows them on standalone text only, per the discussion above), so the alt fallback now only considers the icon: `icon` name minus extension, else `"button"`. `imageAltFallback()` shrank to two lines.

_Posted by Claude Code._

src/app/page/component/content-button/content-button.component.ts
```diff
@@ -77,6 +86,32 @@ export class ContentButtonComponent implements OnChanges, OnDestroy {
...
+  private resolveImage(resource: Resource | null): string | null {
+    if (!resource?.name) {
+      return null;
+    }
+    let url = this.pageService.getImageUrl(resource.name);
+    // Try to find image in all attachments
+    if (url === resource.name && !url.includes('http')) {
+      url = this.pageService.findAttachment(resource.name);
+    }
+    return url || null;
```
frett: I think this URL handing might be pulling images from the attachments table and not from the published content location.

frett: this looks like it might be a bigger issue than just this PR.

Currently the pageService looks to populate an image url lookup service from the attachments data. In reality what is meant to happen is that images are loaded out of the CDN in the same way that the page xml is loaded. The images/attachments for a tool get published alongside the xml.

I'm going to create a jira ticket for this issue, for now this logic can stay assuming it's working. The bigger change will need to be a separate PR later

frett: I created https://jira.cru.org/browse/GT-3059 to migrate from the attachments table mechanism of loading images to actually using the published files mechanism that should be in use

wjames111: Okay so I updated to only pull from the published images list rather then the unpublished attachment URL. Is it okay to leave that way?

frett: @wjames111 I think so, `getImageUrl()` is pre-populated by the tool attachments table, as long as it's working in this PR I'm fine with this logic. The followup ticket will address this to load from the correct location since you'll probably want to centralize the correct url resolution logic into `getImageUrl()`

frett (approved):

